PR #13, opened 2020-09-21
Add style properties and ruby annotations to model

### LINKED ISSUE #14: Add support for style properties to the model

### LINKED ISSUE #15: Add support for ruby annotations

palemieux:
Closes #14
Closes #15

### DIFF
--- a/src/main/python/ttconv/imsc/imsc_reader.py
+++ b/src/main/python/ttconv/imsc/imsc_reader.py
@@ -280,7 +280,10 @@ class ContentElement:
     for attr in ttml_element.attrib:
       prop = StyleProperties.BY_QNAME.get(attr)
       if prop is not None:
-        element.set_style(prop.model_prop, prop.extract(context, ttml_element.attrib.get(attr)))
+        try:
+          element.set_style(prop.model_prop, prop.extract(context, ttml_element.attrib.get(attr)))
+        except ValueError:
+          LOGGER.error("Error reading style property: %s", prop.__name__)
 
   @staticmethod
   def process(context, inherited_space, inherited_lang, ttml_element):
--- a/src/main/python/ttconv/imsc/imsc_reader.py
+++ b/src/main/python/ttconv/imsc/imsc_reader.py
@@ -961,9 +961,7 @@ class CellResolutionAttribute:
 
         return model.CellResolutionType(int(m.group(1)), int(m.group(2)))
 
-      else:
-
-        LOGGER.error("ttp:cellResolution invalid syntax")
+      LOGGER.error("ttp:cellResolution invalid syntax")
 
     # default value in TTML
 
--- a/src/main/python/ttconv/imsc/imsc_reader.py
+++ b/src/main/python/ttconv/imsc/imsc_reader.py
@@ -302,7 +302,35 @@ class ContentElement:
 
     elif ttml_element.tag == SpanElement.qn:
 
-      element = SpanElement.process(context, inherited_space, inherited_lang, ttml_element)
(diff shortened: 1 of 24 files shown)


src/main/python/ttconv/imsc/imsc_reader.py
```diff
@@ -476,90 +530,353 @@ def process(context, inherited_space, inherited_lang, ttml_element):
...
-        element.push_child(model.Text(context.doc, ttml_element.text))
+      if ttml_child_element.tail:
+        element.push_child(model.Text(context.doc, ttml_child_element.tail))
 
     return element
 
-class BrElement:
-  '''Process the TTML <br> element
+
+class RubyElement:
```
valnoel, line 541: In order to limit the size of the files, shouldn't the Ruby-related class definitions be placed into a specific file? And it would be easier to review :wink:

src/main/python/ttconv/imsc/style_properties.py
```diff
@@ -0,0 +1,689 @@
...
+    f"{{{v.ns}}}{v.local_name}" : v
+    for n, v in list(locals().items()) if inspect.isclass(v)
+    }
+
+  @classmethod
+  def ttml_length_to_model(cls, _context: StyleParsingContext, xml_attrib: str):
+    (value, units) = utils.parse_length(xml_attrib)
+
+    return styles.LengthType(value, styles.LengthType.Units(units))
+
```
valnoel, line 689: It seems Git expects a new line here

src/main/python/ttconv/imsc/utils.py
```diff
@@ -0,0 +1,114 @@
...
+        int(m.group(1)),
+        int(m.group(2)),
+        int(m.group(3)),
+        int(m.group(4))
+      ]
+    )
+
+  raise ValueError("Bad Syntax")
+
+_LENGTH_RE = re.compile(r"^((?:\+|\-)?\d*(?:\.\d+)?)(px|em|c|%|rh|rw)$")
```
valnoel, line 85: For readability, I'd rather place the constants at the top of the file, don't you think?

src/main/python/ttconv/imsc/imsc_reader.py
palemieux: I think all element processing should be either kept in the same file (since they are all so similar and dependent) or create one file per element -- there is nothing really special about ruby element processing other than, by habit, I worked on them last :) 
There *is* a lot of code duplication across element processing. Perhaps it is possible to reduce it. Processing is however both annoyingly similar and but also slightly different across elements.

src/main/python/ttconv/imsc/style_properties.py
palemieux: @valnoel Git? Do you mean pylint?

src/main/python/ttconv/imsc/utils.py
palemieux: Fixed at https://github.com/sandflow/ttconv/pull/13/commits/01e3d293233ba407d05f310bda54fd02c0fd19f8

src/main/python/ttconv/imsc/style_properties.py
palemieux: Fixed at https://github.com/sandflow/ttconv/pull/13/commits/c435c36807900c72c2042313ef7749680b71f8b2

src/main/python/ttconv/imsc/imsc_reader.py
valnoel: OK then, we'll see along the project progress if we can simplify all this!

valnoel (approved):

PR #38, opened 2020-09-29
Mapping SCC to model

### LINKED ISSUE #16: Map the SCC content to the main model

valnoel:
Closes #16


src/main/python/ttconv/scc/preambles_access_codes.py
```diff
@@ -24,146 +24,162 @@
...
+      return NamedColors.cyan
+
+    if self._bits in [0x08, 0x09]:
+      return NamedColors.red
+
+    if self._bits in [0x0A, 0x0B]:
+      return NamedColors.yellow
+
+    if self._bits in [0x0C, 0x0D]:
+      return NamedColors.yellow
```
palemieux, line 74: `magenta` right?

src/main/python/ttconv/scc/mid_row_codes.py
```diff
@@ -56,14 +60,53 @@ def get_name(self) -> str:
...
 
+  def get_color(self) -> typing.Optional[NamedColors]:
+    """Returns SCC Mid-Row Code color"""
+    style_bits = self._channel_1 & 0x000F
+
+    if style_bits in [0x00, 0x01]:
+      return NamedColors.white
+    if style_bits in [0x02, 0x03]:
+      return NamedColors.green
+    if style_bits in [0x04, 0x05]:
```
palemieux, line 71: I would use tuples, e.g. `(0x04, 0x05)`, instead of lists since tuples are immutable.

palemieux: Using tuples would allow the use a dictionary instead of `if...return` statements, e.g.

```
mapping = {
(0x00, 0x01) : NamedColors.white
}
```

src/main/python/ttconv/scc/preambles_access_codes.py
valnoel: Fixed in 681cd3a5e52267ab5ca24850bde5bc52d975cbc6

src/main/python/ttconv/scc/mid_row_codes.py
valnoel: Done in 681cd3a5e52267ab5ca24850bde5bc52d975cbc6

PR #46, opened 2020-10-02
SCC time mapping

### LINKED ISSUE #45: SCC time mapping
Extract time information from SCC data stream, and map it to the data model

valnoel:
Closes #45


palemieux (changes requested):
For drop frame timecode, _the first two frame numbers (00 and 01) shall be omitted from the count at the start of each minute except minutes 00, 10, 20, 30, 40, and 50._

src/test/python/test_scc_time_codes.py
```diff
@@ -51,32 +52,227 @@ def test_parse_drop_frame_time_code(self):
     self.assertEqual(3, time_code.get_seconds())
     self.assertEqual(4, time_code.get_frames())
     self.assertTrue(time_code.is_drop_frame())
-    self.assertEqual(111582, int(time_code._get_frames()))
-    self.assertAlmostEqual(3719.414, float(time_code.get_fraction()), delta=0.001)
+    self.assertEqual(111582, time_code.get_nb_frames())
+    self.assertAlmostEqual(111582 / 30, float(time_code.get_fraction()), delta=0.001)
```
palemieux, line 56: shouldn't this be:

```self.assertEqual(111582 / Fraction(30000/1001), time_code.get_fraction())```

palemieux (changes requested):

src/test/python/test_scc_time_codes.py
valnoel: No, in my opinion it wouldn't make sense since the number of frames (used as numerator) is absolute. That's why the fraction is computed with the non-drop-frame frame rate as denominator by default.

valnoel: Fixed in https://github.com/sandflow/ttconv/pull/46/commits/2580698102641c728205ac27b125fdb157849b34

palemieux (approved):

PR #52, opened 2020-10-12
Initial IMSC Writer implementation

### LINKED ISSUE #22: Implement IMSC writer
Implement a converter from the model to an [IMSC 1.1](https://www.w3.org/TR/ttml-imsc1.1/) document :

* Complete `ttconv/imsc/imsc_writer.py` to convert a model `Document` into a valid IMSC 1.1 document. There should be one function/method for each `ContentElement` model subclass that converts the subclass to the equivalent XML node.
* add a `write` method to each `StyleProperty` at `ttconv/imsc/style_properties.py` take as input the value of a model style property and outputs an XML attribute
* allow the user to chose whether to output time expressions using either:
  * `HH:MM:SS.ms` (hours-minutes-seconds-milliseconds)
  * `xxxxxxf` (frames)

The objective is to exercise the full capabilities of the model, which is based on IMSC 1.1. In particular, it should be possible to convert an IMSC document to another IMSC document, such that both render to the same result.

### LINKED ISSUE #66: IMSC reader does not set begin/end on discrete animation steps
https://github.com/sandflow/ttconv/blob/692e58180c621cdfe3462239d0203f008a130d96/src/main/python/ttconv/imsc/elements.py#L572

This needs to be called after the temporal attributes are computed.

BobBrownConsulting:
Closes #22
Closes #66


palemieux (approved):

BobBrownConsulting:
Closes #22
Closes #66

PR #56, opened 2020-10-19
SCC Paint-On captions support

### LINKED ISSUE #41: Handle Paint-On SCC caption style

valnoel:
Closes #41


src/main/python/ttconv/scc/scc_reader.py
```diff
@@ -145,24 +159,30 @@ def process_control_code(self, control_code: SccControlCode, time_code: SccTimeC
...
       self.set_current_to_previous()
 
       previous_last_lines: List[SccCaptionContent] = []
-      if self.previous_caption:
+      if len(self.previous_captions) > 0:
 
         if control_code is SccControlCode.RU2:
-          previous_last_lines = self.previous_caption.get_last_caption_lines(1)
+          previous_last_lines = self.previous_captions[0].get_last_caption_lines(1)
         if control_code is SccControlCode.RU3:
```
palemieux, line 178: Why not `elif` here?

palemieux (commented):
https://github.com/sandflow/ttconv/blob/0f013fb727a900c86ed191ff248984704d4d5e68/src/main/python/ttconv/scc/scc_reader.py#L51

Suggest using `os.environ['DEBUG']`

src/main/python/ttconv/scc/scc_reader.py
palemieux: Similarly, why not use `elif` throughout the following function?

https://github.com/sandflow/ttconv/blob/0f013fb727a900c86ed191ff248984704d4d5e68/src/main/python/ttconv/scc/scc_reader.py#L155

palemieux (changes requested):
See comments

src/main/python/ttconv/scc/scc_reader.py
valnoel: Fixed in 14f91c9

src/main/python/ttconv/scc/scc_elements.py
```diff
@@ -194,16 +227,24 @@ def get_origin(self) -> PositionType:
 
   def get_extent(self) -> ExtentType:
     """Computes and returns the current paragraph extent, based on its content"""
-    if len(self.caption_contents):
-      nb_lines = sum(map(lambda br: isinstance(br, SccCaptionLineBreak), self.caption_contents)) + 1
-      text_lengths = [0] + [len(caption_text.text)
-                            for caption_text in self.caption_contents
-                            if isinstance(caption_text, SccCaptionText)]
-      max_text_lengths = max(text_lengths)
+    lines: List[str] = []
```
palemieux, line 230: Why is the extent of the paragraph important? In the model, paragraphs do not have extent. Instead, paragraphs flow into a region, which has an extent. I am not convinced that the size of the region should change dynamically. Instead perhaps only the origin of the region could change over time.

palemieux (changes requested):

palemieux (approved):

PR #87, opened 2020-10-28
 SCC reader output length in percentage units

### LINKED ISSUE #86: SCC reader should not output length in cell units
IMSC prohibits the use of cell units but for `ebutts:linePadding` -- see https://www.w3.org/TR/ttml-imsc1.1/#length-cell

Instead SCC reader should use percentages.

valnoel:
Closes #86

### DIFF
--- a/src/main/python/ttconv/scc/utils.py
+++ b/src/main/python/ttconv/scc/utils.py
@@ -27,6 +27,7 @@
 from numbers import Number
 from typing import Union
 
+from ttconv.model import CellResolutionType
 from ttconv.style_properties import LengthType, PositionType, ExtentType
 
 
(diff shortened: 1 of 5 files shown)


src/main/python/ttconv/scc/utils.py
```diff
@@ -44,3 +45,25 @@ def get_extent_from_dimensions(width: Union[int, Number], height: Union[int, Num
...
+
+
+def convert_cells_to_percentages(dimensions: Union[PositionType, ExtentType], cell_resolution: CellResolutionType) -> Union[
+  PositionType, ExtentType]:
+  """Converts dimensions from cell units to percentage units"""
+  if isinstance(dimensions, PositionType):
+    if dimensions.x.units is not LengthType.Units.c or dimensions.y.units is not LengthType.Units.c:
+      raise ValueError("Dimensions to convert must be in cell units")
+
+    x_percent = dimensions.x.value * 100 / cell_resolution.columns
```
palemieux, line 57: Percentages should be rounded to the nearest integer.

palemieux (approved):

src/main/python/ttconv/scc/utils.py
```diff
@@ -44,3 +45,25 @@ def get_extent_from_dimensions(width: Union[int, Number], height: Union[int, Num
...
+    y_percent = dimensions.y.value * 100 / cell_resolution.rows
```
palemieux, line 58: Percentages should be rounded to the nearest integer

palemieux (changes requested):

src/main/python/ttconv/scc/utils.py
valnoel: Fixed in 17ee6eb

src/main/python/ttconv/scc/utils.py
valnoel: Fixed in 17ee6eb

palemieux (approved):

PR #91, opened 2020-10-28
Update TT convert to support SCC to IMSC conversion

### LINKED ISSUE #74: SCC to IMSC converter CLI
Implement support for conversion from SCC to IMSC using the `tt` CLI, i.e.

`tt convert --from sample.scc --to sample.ttml`

Include a system unit test using the sample SCC file at https://labs.inqscribe.com/exporter/scc/.

palemieux:
Closes #74


src/main/python/ttconv/imsc/attributes.py
```diff
@@ -133,7 +133,7 @@ def extract(ttml_element) -> model.CellResolutionType:
 
   @staticmethod
   def set(ttml_element, res):
```
BobBrownConsulting, line 135: Where does the writing of px happen now?

BobBrownConsulting (approved):

src/main/python/ttconv/imsc/attributes.py
palemieux: Cell resolution is in number of row and columns, so no units.

See https://www.w3.org/TR/ttml2/#parameter-attribute-cellResolution

PR #123, opened 2020-11-09
Implement ISD style inheritance and computation

### LINKED ISSUE #115: Handle tts:textDecoration inheritance and specification
https://github.com/sandflow/imscJS/blob/6039c0ab2fe89c4ae35332aa438441e3b06d580f/src/main/js/isd.js#L189

### LINKED ISSUE #116: Handle tts:fontSize inheritance for Rt
https://github.com/sandflow/imscJS/blob/6039c0ab2fe89c4ae35332aa438441e3b06d580f/src/main/js/isd.js#L240

### LINKED ISSUE #121: Handle style computation during inheritance processing
Style property computation, e.g. `fontSize` expressed in `%`, is not implemented

### LINKED ISSUE #122: The TextDecorationType model is incorrect
`"none"` is equivalent to `underline == False and line_through== False and overline == False`

`TextDecorationType` should be replaced with:

```
@dataclass(frozen=True)
class TextDecorationType:
  underline: typing.Optional[bool] = None
  line_through: typing.Optional[bool] = None
  overline: typing.Optional[bool] = None
```

### LINKED ISSUE #124: TextEmphasis model is incorrect
The `TextEmphasis` class should be refactored to remove `none` from the list of possible symbols since `none` implies no color and no position.

### LINKED ISSUE #126: ISD Text elements should contain text

### LINKED ISSUE #127: ISD for element end time should not contain the element
For instance, for the following TTML paragraph:
```xml
<p region="r1" begin="0s" end="1s">Region with tts:origin="5% 5%" and tts:extent="80% 20%"</p>
<p region="r2" begin="1s" end="2s">Region with tts:origin="5% 5%" and tts:extent="80% 20%"</p>
```
the first generated ISD (for the 0s offset) contains the first paragraph content. whereas the second ISD (for the 1s offset) contains the both paragraphs.

This second ISD should not contain the first paragraph.

### LINKED ISSUE #129: Sort returned ISD significant times
Sort the Fractions set returned by `ISD.significant_times(doc)`

palemieux:
Closes #129 
Closes #121 
Closes #122 
Closes #115 
Closes #116
Closes #124 
Closes #126 
Closes #127

### DIFF
--- a/src/main/python/ttconv/isd.py
+++ b/src/main/python/ttconv/isd.py
@@ -29,6 +29,7 @@ from __future__ import annotations
 
 import inspect
 import typing
+import numbers
 from fractions import Fraction
 
 import ttconv.model as model
@@ -179,6 +180,9 @@ class ISD(model.Root):
     if styles.StyleProperties.FontSize in styles_to_be_computed:
       StyleProcessors.BY_STYLE_PROP[styles.StyleProperties.FontSize].compute(isd_parent, isd_element)
 
+    if styles.StyleProperties.Extent in styles_to_be_computed:
+      StyleProcessors.BY_STYLE_PROP[styles.StyleProperties.Extent].compute(isd_parent, isd_element)
+
   @staticmethod
   def process_element(isd: ISD,
                       offset: Fraction,
@@ -187,6 +191,8 @@ class ISD(model.Root):
                       parent: typing.Optional[model.ContentElement],
                       element: model.ContentElement
                       ) -> typing.Optional[model.ContentElement]:
+    if element is None:
+      return None
 
     doc = element.get_doc()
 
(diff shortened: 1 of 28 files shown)


src/main/python/ttconv/isd.py
```diff
@@ -80,7 +81,7 @@ def get_region(self, region_id) -> typing.Optional[ISD.Region]:
 
   def iter_regions(self) -> typing.Iterator[ISD.Region]:
     '''Returns an iterator over regions.'''
-    return self._regions.values()
+    return self._regions.values()
```
valnoel, line 92: ⚠️ Trailing whitespace

src/main/python/ttconv/isd.py
```diff
@@ -340,13 +343,40 @@ def process_element(isd: ISD,
...
+            raise ValueError("Attempting to compute a relative length for style"
+                             f" property {cls.style_prop.__name__} whose parent is None")
+          style_value = styles.LengthType(
+            value=parent_length.value * style_value.value / 100,
+            units=parent_length.units
+          )
+        else:
+          style_value = styles.LengthType(
+            value=style_value.value / 100,
+            units=styles.LengthType.Units.c
```
valnoel: Shouldn't we compare with the root cell resolution here?

valnoel (changes requested):

valnoel (approved):

PR #128, opened 2020-11-12
ISD: Fixes output Text and Region content

### LINKED ISSUE #126: ISD Text elements should contain text

### LINKED ISSUE #127: ISD for element end time should not contain the element

valnoel:
Closes #126 
Closes #127

### DIFF
--- a/src/main/python/ttconv/isd.py
+++ b/src/main/python/ttconv/isd.py
@@ -222,6 +222,9 @@ class ISD(model.Root):
       isd_element.set_lang(element.get_lang())
       isd_element.set_space(element.get_space())
 
+    if isinstance(element, model.Text):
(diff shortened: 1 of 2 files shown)


palemieux:
Combined in #123

PR #131, opened 2020-11-17
Add unit test coverage to CI script

### LINKED ISSUE #64: Add coverage to CI
CI should fail if coverage is less than 80%

palemieux:
Closes #64


src/main/python/ttconv/imsc/style_properties.py
```diff
@@ -55,7 +55,7 @@ def local_name(cls):
     raise NotImplementedError
 
   @classmethod
```
BobBrownConsulting, line 57: Fixing pylint unused variable

src/test/python/test_imsc_reader.py
```diff
@@ -35,9 +35,52 @@
 import ttconv.model as model
 import ttconv.style_properties as styles
 import ttconv.imsc.reader as imsc_reader
+import ttconv.imsc.namespaces as xml_ns
 
 class IMSCReaderTest(unittest.TestCase):
```
BobBrownConsulting, line 41: Not certain if this is needed. Happy to remove if not.

src/test/python/test_imsc_reader.py
```diff
@@ -35,9 +35,52 @@
...
 
+  def setUp(self):
+    et.register_namespace("ttml", xml_ns.TTML)
+    et.register_namespace("ttp", xml_ns.TTP)
+    et.register_namespace("tts", xml_ns.TTS)
+    et.register_namespace("ittp", xml_ns.ITTP)
+    et.register_namespace("itts", xml_ns.ITTS)
+
+  def test_reader_tt_element_not_root_element(self):
+
```
BobBrownConsulting, line 50: Covers first early return from reader.

src/test/python/test_imsc_reader.py
```diff
@@ -35,9 +35,52 @@
...
+          xmlns:ttm="http://www.w3.org/ns/ttml#metadata" 
+          xmlns:tts="http://www.w3.org/ns/ttml#styling"
+          xmlns:ttp="http://www.w3.org/ns/ttml#parameter" 
+          xmlns:ittp="http://www.w3.org/ns/ttml/profile/imsc1#parameter"
+          ittp:activeArea="50% 50% 80% 80%"
+          tts:extent="640px 480px"
+          ttp:profile="http://www.w3.org/ns/ttml/profile/imsc1/text">
+          <asdf></asdf>
+      </tt>"""
+
```
BobBrownConsulting, line 80: This XML fails as <asdf> is not a valid TT value. The reader throws. 
Is the second None return in reader even valid?
What XML would trigger that case?
It seems the reader throws in many cases of invalid XML.

palemieux: Replace `<tt>` with `<foo>`, i.e. `None` will be returned if the root element is not `<tt>`

palemieux: Oh... this is already covered by `test_reader_tt_element_not_root_element`. I suggest removing this test.

src/test/python/test_imsc_reader.py
palemieux: What are you referring to?

BobBrownConsulting: def setUp(self):

src/test/python/test_imsc_reader.py
BobBrownConsulting: yes, that is correct. I wasn't sure if there was another way to pass the first check in to_model while still triggering the second check. I tried several things but the to_model alway threw exceptions which is the right way to do it.

BobBrownConsulting (approved):

src/test/python/test_imsc_reader.py
palemieux: Already defined in `imsc_writer`

PR #138, opened 2020-11-20
Strip whitespace during ISD construction

### LINKED ISSUE #117: Should ISD processing trim whitespace during ISD construction?
https://github.com/sandflow/imscJS/blob/6039c0ab2fe89c4ae35332aa438441e3b06d580f/src/main/js/isd.js#L512

This may result in empty Text that needs to be pruned.

palemieux:
Closes #117

### DIFF
--- a/src/main/python/ttconv/isd.py
+++ b/src/main/python/ttconv/isd.py
@@ -30,6 +30,7 @@ from __future__ import annotations
 import inspect
 import typing
 import numbers
+import re
 from fractions import Fraction
 
 import ttconv.model as model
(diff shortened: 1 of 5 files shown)


src/main/python/ttconv/isd.py
```diff
@@ -397,6 +404,64 @@ def _process_element(
 
     return None
 
+def _prune_empty_spans(element: model.ContentElement):
+  children = list(element)
+  for child in children:
+    _prune_empty_spans(child)
+    if isinstance(child, model.Text) and not child.get_text():
+      element.remove_child(child)
+    elif isinstance(child, model.Span) and not child:
```
valnoel, line 413: What do you mean by `not child` here? It cannot be `None` and instance of `Span`, right?

valnoel (approved):

PR #143, opened 2020-11-20
Improve IMSC time syntax output

### LINKED ISSUE #102: IMSC Writer should allow the user to select a frame rate or tick rate
IMSC writer currently output all temporal offsets in `s` units.

The user should be able to select output as:
* `hh:mm:ss.ms`
* `FFFFf` where the user would specify a frame rate

For example,

`begin="0.7674333333333333s"`

would become:

`begin="30f"` given a frame rate of `30000/1001`, i.e. 29.97.

BobBrownConsulting:
Closes #102


src/main/python/ttconv/imsc/attributes.py
```diff
@@ -297,6 +298,31 @@ class TemporalAttributeParsingContext:
...
+  value = time.numerator / time.denominator
+  value = datetime.timedelta(seconds = value)
+  #millsec = value.microseconds / 1000
+  #value = str(datetime.timedelta(seconds = value)) 
+
+  return f"{value}s"
+
+def to_frames(context: TemporalAttributeWritingContext, time: Fraction) -> str:
+  value = time.numerator / time.denominator
+  value *= (context.frame_rate.numerator / context.frame_rate.denominator)
```
BobBrownConsulting, line 321: Currently truncating. This is likely the correct behavior as if you are not on a frame boundary, you are in a frame.

src/main/python/ttconv/imsc/attributes.py
```diff
@@ -297,6 +298,31 @@ class TemporalAttributeParsingContext:
...
+  tick_rate: int = 1
+
+def to_time_format(context: TemporalAttributeWritingContext, time: Fraction) -> str:
+  if context.frame_rate.numerator == 1 and context.frame_rate.denominator == 1:
+    return to_hh_mm_ss_ms(time)
+  return to_frames(context, time)
+
+def to_hh_mm_ss_ms(time: Fraction) -> str:
+  value = time.numerator / time.denominator
+  value = datetime.timedelta(seconds = value)
```
BobBrownConsulting, line 313: Currently returns microseconds in the last field. Do we want milliseconds?

palemieux: Yes, we should round to 3 digits, e.g. `round(x, 3)`

src/main/python/ttconv/imsc/attributes.py
palemieux: `Fraction` instance should be multiplied, so that precision can be preserved.

See https://www.w3.org/TR/ttml-imsc1.1/#synchronization for trunc vs ceiling

palemieux (changes requested):

src/main/python/ttconv/imsc/elements.py
```diff
@@ -191,7 +199,8 @@ def from_xml(_parent_ctx: typing.Optional[TTMLElement.ParsingContext], xml_elem)
     return tt_ctx
 
   @staticmethod
-  def from_model(model_doc: model.Document) -> et.Element:
+  def from_model(model_doc: model.Document, frame_rate = Fraction(1,1)) -> et.Element:
```
palemieux, line 202: Can `frame_rate` be `None`? How would the caller specify that conversion to `f` is not desired?

BobBrownConsulting: Done. None means use hh:mm:ss:ms otherwise frames

src/main/python/ttconv/imsc/attributes.py
BobBrownConsulting: Done

src/main/python/ttconv/imsc/attributes.py
BobBrownConsulting: Done

src/main/python/ttconv/imsc/attributes.py
```diff
@@ -297,6 +299,26 @@ class TemporalAttributeParsingContext:
...
+def to_time_format(context: TemporalAttributeWritingContext, time: Fraction) -> str:
+  if context.frame_rate is None:
+    return to_hh_mm_ss_ms(time)
+  return to_frames(context, time)
+
+def to_hh_mm_ss_ms(time: Fraction) -> str:
+  millisecond = (time % 1) * 1000
+  minute, second = divmod(time, 60) 
+  hour, minute = divmod(minute, 60) 
+  return "%d:%02d:%02d.%03ds" % (hour, minute, second, millisecond)
```
palemieux, line 315:
```suggestion
  return f"%d:%02d:%02d.%03ds"
```

src/main/python/ttconv/imsc/attributes.py
```diff
@@ -297,6 +299,26 @@ class TemporalAttributeParsingContext:
...
+  frame_rate: Fraction = Fraction(30, 1)
+
+def to_time_format(context: TemporalAttributeWritingContext, time: Fraction) -> str:
+  if context.frame_rate is None:
+    return to_hh_mm_ss_ms(time)
+  return to_frames(context, time)
+
+def to_hh_mm_ss_ms(time: Fraction) -> str:
+  millisecond = (time % 1) * 1000
+  minute, second = divmod(time, 60)
```
palemieux, line 313: @valnoel Don't we have some code somewhere that converts from seconds to HH:MM:SS?

src/test/python/test_imsc_writer.py
```diff
@@ -595,5 +603,42 @@ def test_style_property_TextShadow_has_px(self):
...
+    self.assertEqual(val, '90f')
+
+    context.frame_rate = Fraction(25000, 1000)
+    time = Fraction(3600, 1)
+    val = attributes.to_time_format(context, time)
+    self.assertEqual(val, '90000f')
+
+    context.frame_rate = Fraction(30000, 1001)
+    time = float(0.7674333333333333).as_integer_ratio()
+    time = Fraction(time[0], time[1])
```
palemieux, line 639:
```suggestion
    time = Fraction(0.7674333333333333)
```

palemieux (changes requested):

src/test/python/test_imsc_writer.py
```diff
@@ -595,5 +603,42 @@ def test_style_property_TextShadow_has_px(self):
...
+    val = attributes.to_time_format(context, time)
+    self.assertEqual(val, '23f')
+
```
palemieux, line 642:
```suggestion
    context.frame_rate = Fraction(30000, 1001)
    time = 2301 / Fraction(30000, 1001)
    val = attributes.to_time_format(context, time)
    self.assertEqual(val, '2301f')
```

src/main/python/ttconv/imsc/attributes.py
valnoel: Yes, we have such code in the [SRT module](https://github.com/sandflow/ttconv/blob/master/src/main/python/ttconv/srt/time_code.py#L45), and to convert frames to time code as well in the [SCC module](https://github.com/sandflow/ttconv/blob/master/src/main/python/ttconv/scc/time_codes.py#L81).

Isn't it time to merge all this into a common module?

BobBrownConsulting: What should we do here?
Move the SRT implementation to a generic one and then have the SRT implementation call the generic and the wrap the time into SrtTimeCode

valnoel: In my opinion, if we create a generic time code implementation that handle conversions from/to frames and seconds, we could simply remove the SrtTimeCode and SccTimeCode classes.
But I think we should create a specific issue for this refactoring, don't you think?

BobBrownConsulting:
@palemieux I've merged master to the branch and resolved conflicts.

src/main/python/ttconv/imsc/attributes.py
BobBrownConsulting: Based on conversation at standup, we are not going to fix this as part of this PR. 
Logged issue https://github.com/sandflow/ttconv/issues/161

src/main/python/ttconv/imsc/attributes.py
```diff
@@ -297,6 +299,26 @@ class TemporalAttributeParsingContext:
...
+def to_time_format(context: TemporalAttributeWritingContext, time: Fraction) -> str:
+  if context.frame_rate is None:
+    return to_hh_mm_ss_ms(time)
+  return to_frames(context, time)
+
+def to_hh_mm_ss_ms(time: Fraction) -> str:
+  millisecond = (time % 1) * 1000
+  minute, second = divmod(time, 60) 
+  hour, minute = divmod(minute, 60) 
+  return f"{hour}:{minute:02}:{int(second):02}.{int(millisecond):03}s"
```
palemieux, line 315: The format is `HH:MM:SS:sss` with no `"s"` at the end

palemieux (changes requested):

BobBrownConsulting:
@palemieux I fixed the issue with the "s". 

I also sync up with master which changes the interfaces to imsc.from_model as it takes both a frame rate and progress bar. Please double check that implementation to make sure it is OK with you.

CI passes

palemieux:
> I also sync up with master which changes the interfaces to imsc.from_model as it takes both a frame rate and progress bar. Please double check that implementation to make sure it is OK with you.

@valnoel @BobBrownConsulting I am thinking we need come up with a more general way of passing parameters to the readers and writers, e.g. through a config file and command line params. Thoughts?

src/main/python/ttconv/imsc/attributes.py
```diff
@@ -297,6 +299,26 @@ class TemporalAttributeParsingContext:
...
+def to_time_format(context: TemporalAttributeWritingContext, time: Fraction) -> str:
+  if context.frame_rate is None:
+    return to_hh_mm_ss_ms(time)
+  return to_frames(context, time)
+
+def to_hh_mm_ss_ms(time: Fraction) -> str:
+  millisecond = (time % 1) * 1000
+  minute, second = divmod(time, 60) 
+  hour, minute = divmod(minute, 60) 
+  return f"{hour}:{minute:02}:{int(second):02}.{int(millisecond):03}"
```
valnoel, line 315: Shouldn't hours be not zero-padded as well?

src/main/python/ttconv/imsc/elements.py
```diff
@@ -89,6 +89,14 @@ def process_space_attribute(self, parent_ctx: TTMLElement.ParsingContext, xml_el
       space_attr_value = imsc_attr.XMLSpaceAttribute.extract(xml_elem)
       self.space = space_attr_value if space_attr_value is not None else parent_ctx.space
 
+  class WritingContext:
+    '''State information when writing a TTML element'''
+
+    def __init__(self, frame_rate: imsc_attr.TemporalAttributeWritingContext):
```
valnoel, line 95: `frame_rate` should be a `Fraction` instance, isn't it?

src/main/python/ttconv/imsc/attributes.py
```diff
@@ -297,6 +299,26 @@ class TemporalAttributeParsingContext:
   frame_rate: Fraction = Fraction(30, 1)
   tick_rate: int = 1
 
+@dataclass
+class TemporalAttributeWritingContext:
+  frame_rate: Fraction = Fraction(30, 1)
```
valnoel, line 304: Wouldn't it be simplier to initialize this `frame_rate` attribute into a `__init__` with a default `Fraction` value?

src/main/python/ttconv/imsc/elements.py
BobBrownConsulting: Good catch. Yes it should.

src/main/python/ttconv/imsc/attributes.py
BobBrownConsulting: @dataclass automatically adds the __init__ 

https://docs.python.org/3/library/dataclasses.html

I'm not sure if adding our own __init__ would make this more clear or not.

@palemieux Do you have any thoughts on this?

src/main/python/ttconv/imsc/elements.py
```diff
@@ -89,6 +89,13 @@ def process_space_attribute(self, parent_ctx: TTMLElement.ParsingContext, xml_el
       space_attr_value = imsc_attr.XMLSpaceAttribute.extract(xml_elem)
       self.space = space_attr_value if space_attr_value is not None else parent_ctx.space
 
+  class WritingContext:
+    '''State information when writing a TTML element'''
+
+    def __init__(self, frame_rate: Fraction):
+      self.temporal_context = imsc_attr.TemporalAttributeWritingContext
+      self.temporal_context.frame_rate = frame_rate
```
valnoel, line 97:
```python
self.temporal_context = imsc_attr.TemporalAttributeWritingContext(frame_rate)
```
:)

valnoel (changes requested):

palemieux (approved):

PR #146, opened 2020-11-23
Fixes CI unit tests

valnoel:


palemieux (changes requested):
IMSC 1.2 is not supported. Can we drop `test_imsc_1_2_test_suite`?

palemieux (approved):

Issue #145, opened 2020-11-23
Add progress bar to tt CLI


palemieux:
Maybe consider a callback handler on `from_model` and `to_model` for writers and readers.

BobBrownConsulting:
@palemieux @valnoel The core of this is in place. The next step is to have both of you review where the progress_callbacks are inserted and add/remove them and provide more information about current progress (putting 0 to 1 in as the percent_progress)

PR #148, opened 2020-11-24
Adding progress callback to provide feedback by writing dots . to the…

BobBrownConsulting:
… command line (#145))

### DIFF
--- a/src/main/python/ttconv/imsc/elements.py
+++ b/src/main/python/ttconv/imsc/elements.py
@@ -96,7 +96,7 @@ class TTMLElement:
     raise NotImplementedError
 
   @staticmethod
-  def from_xml(parent_ctx, xml_elem, progress_callback = None):
+  def from_xml(parent_ctx, xml_elem, progress_callback=lambda _: None):
     '''Returns a parsing context for the TTML element represented by the XML element `xml_elem` and
     given the parent context `parent_ctx`
     '''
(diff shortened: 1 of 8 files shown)


src/main/python/ttconv/tt.py
```diff
@@ -39,6 +39,10 @@
 
 LOGGER = logging.getLogger(__name__)
```
BobBrownConsulting, line 41: Writing out the "." should probably be throttled by time or something else such that we don't fill up our terminal

palemieux: What about the callback returning a value between 0 and 1?

palemieux (changes requested):

src/main/python/ttconv/tt.py
BobBrownConsulting: Done

src/main/python/ttconv/imsc/elements.py
```diff
@@ -96,7 +96,7 @@ def is_instance(xml_elem) -> bool:
     raise NotImplementedError
 
   @staticmethod
-  def from_xml(parent_ctx, xml_elem):
+  def from_xml(parent_ctx, xml_elem, progress_callback = None):
```
palemieux, line 99: Any reason for not moving `progress_callback` into `parent_ctx`?

valnoel (commented):
@BobBrownConsulting @palemieux 
I'd suggest that the default `progress_callback` parameter value to be an empty function, so that we don't have to check whether the callback is set or None.
Something like:
```python
def to_model(..., progress_callback=lambda _: None):
    ...
    progress_callback(0.5)
    ...
```
instead of:
```python
def to_model(..., progress_callback=None):
    ...
    if progress_callback:
        progress_callback(0.5)
    ...
```

src/main/python/ttconv/imsc/elements.py
BobBrownConsulting: I don't think putting this in the parent_ctx is the right thing to do. There isn't the same parallel for the writer (from_model) 

Also, the parent context is currently owned/created in class TTElement(TTMLElement): from_xml. 

The caller who pass in the progress bar is the tt.py layer which passes it into reader to_model which passes in a None for parent context via the call imsc_elements.TTElement.from_xml

Give that, making the parent context own it, seems odd to me.

BobBrownConsulting:
@palemieux I've updated to master and resolve conflicts. Please do a final pass for the review.

palemieux (approved):

PR #157, opened 2020-12-02
Improve ISD creation performance

### LINKED ISSUE #156: Improve ISD creation performance
Option 1: Parallelize the creation of ISDs across significant times using https://docs.python.org/3/library/multiprocessing.html

Option 2: Use `map` instead of `for` loops

palemieux:
* change inner for loop to functional loop

Closes #156


palemieux:
Replaced by #160

PR #172, opened 2020-12-14
Issues/163 logging interferes with progress bar

### LINKED ISSUE #163: Logging output interferes with tt.py progress bar
image

BobBrownConsulting:
Closes #163

### DIFF
--- a/src/main/python/ttconv/tt.py
+++ b/src/main/python/ttconv/tt.py
@@ -270,14 +270,14 @@ def validate(args):
   print(args.input)
   LOGGER.info("Input file is %s", args.input)
 
+# Ensure that the handler is added only once/globally
+# Otherwise the handler will be called multiple times
+progress = ProgressConsoleHandler()
+LOGGER.addHandler(progress)
+LOGGER.setLevel(logging.INFO) 
 
 def main(argv):
   '''Main application processing'''
-
-  LOGGER.handlers.clear()
-  progress = ProgressConsoleHandler()
-  LOGGER.setLevel(logging.INFO) 
-  LOGGER.addHandler(progress)
 
   args = cli.parse_args(argv)
   if args.subcommand is None:
--- a/src/main/python/ttconv/tt.py
+++ b/src/main/python/ttconv/tt.py
@@ -49,6 +49,21 @@ class ProgressConsoleHandler(logging.StreamHandler):
   is_writing_progress_bar = False
   last_progress_msg = ""
 
+  def progress_str(self, percent_progress: float) -> str:
+    '''Formats the progress string.'''
+
+    total = 1.0
+    prefix = "Reading Progress:" if READING else "Writing Progress:"
+    suffix = 'Complete'
+    decimals = 1
+    length = 50
+    fill = '█'
+    percent = ("{0:." + str(decimals) + "f}").format(100 * (percent_progress / float(total)))
+    filled_length = int(length * percent_progress // total)
+    bar_val = fill * filled_length + '-' * (length - filled_length)
+    
+    return f'\r{prefix} |{bar_val}| {percent}% {suffix}'
+
   def emit(self, record):
     try:
       msg = self.format(record)
@@ -56,19 +71,17 @@ class ProgressConsoleHandler(logging.StreamHandler):
       stream = self.stream
       writing_progress_bar = hasattr(record, 'writing_progress_bar')
 
-      percent = None
+      percent_progress = None
       if writing_progress_bar:
+        percent_progress = getattr(record, 'percent_progress')
+        msg = self.progress_str(float(percent_progress))
         self.last_progress_msg = msg
-        percent = getattr(record, 'percent')
   
       if self.is_writing_progress_bar and not writing_progress_bar:
         # erase and over write the progress bar
         stream.write('\r')
         length = len(self.last_progress_msg)
-        i = 0
-        while i < length:
-          stream.write(' ')
-          i += 1
+        stream.write(' ' * length)
 
         # go to beginning of the line and write the new messages
         stream.write('\r')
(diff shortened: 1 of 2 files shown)


src/main/python/ttconv/tt.py
```diff
@@ -225,6 +270,11 @@ def validate(args):
   print(args.input)
   LOGGER.info("Input file is %s", args.input)
 
+# Ensure that the handler is added only once/globally
+# Otherwise the handler will be called multiple times
+progress = ProgressConsoleHandler()
```
palemieux, line 275: No need to define the `progress` module variable

src/main/python/ttconv/tt.py
```diff
@@ -225,6 +270,11 @@ def validate(args):
...
+LOGGER.addHandler(progress)
```
palemieux, line 276: ``
```suggestion
LOGGER.addHandler(ProgressConsoleHandler())
```

src/test/python/test_progress_logging.py
```diff
@@ -0,0 +1,57 @@
...
+    LOGGER.info('test 2')
+    LOGGER.info('test 3')
+    tt.progress_callback(0.1)
+    tt.progress_callback(1)
+    LOGGER.info('test 4')
+    tt.progress_callback(0.3)
+    LOGGER.info('test 5')
+    tt.LOGGER.info('test 6')
+    tt.progress_callback(1)
+    LOGGER.info('test 7')
```
palemieux, line 54: Suggest capturing the log events with https://docs.python.org/3.7/library/unittest.html#unittest.TestCase.assertLogs and making sure they come out in the expected order

src/main/python/ttconv/tt.py
```diff
@@ -37,36 +37,81 @@
...
+  one line for selected messages
   """
-  percent = ("{0:." + str(decimals) + "f}").format(100 * (iteration / float(total)))
-  filled_length = int(length * iteration // total)
-  bar_val = fill * filled_length + '-' * (length - filled_length)
-  print(f'\r{prefix} |{bar_val}| {percent}% {suffix}', end = print_end)
-  # Print New Line on Complete
-  if iteration == total: 
-    print()
+  is_writing_progress_bar = False
```
palemieux, line 49: This should be set in the `__init__` method otherwise it is shared by all instances of the class.

src/main/python/ttconv/tt.py
```diff
@@ -37,36 +37,81 @@
...
+  last_progress_msg = ""
```
palemieux, line 50: This should be set in the `__init__` method otherwise it is shared by all instances of the class.

src/main/python/ttconv/tt.py
```diff
@@ -37,36 +37,81 @@
...
+    except (KeyboardInterrupt, SystemExit):
+      raise
+    except:
+      self.handleError(record)
 
 def progress_callback(percent_progress: float):
   '''Callback handler used by reader and writer.'''
-  prefix_str = "Reading Progress:" if READING else "Writing Progress:"
-  print_progress_bar(percent_progress, 1.0, prefix = prefix_str, suffix = 'Complete', length = 50)
+  LOGGER.info('', extra={'writing_progress_bar':True, 'percent_progress':percent_progress})
```
palemieux, line 114: Suggest adding a log message even it is ignored by the handler.
```suggestion
  LOGGER.info(f'{percent_progress}%', extra={'writing_progress_bar':True, 'percent_progress':percent_progress})
```

src/main/python/ttconv/tt.py
```diff
@@ -37,36 +37,81 @@
...
-  if iteration == total: 
-    print()
+  is_writing_progress_bar = False
+  last_progress_msg = ""
+
+  def progress_str(self, percent_progress: float) -> str:
+    '''Formats the progress string.'''
+
+    total = 1.0
+    prefix = "Reading Progress:" if READING else "Writing Progress:"
```
palemieux, line 56: Shouldn't `READING` be a field of the handler instead of a global?

palemieux (changes requested):

src/main/python/ttconv/tt.py
BobBrownConsulting: fixed

src/main/python/ttconv/tt.py
BobBrownConsulting: fixed

src/main/python/ttconv/tt.py
BobBrownConsulting: Fixed

src/main/python/ttconv/tt.py
BobBrownConsulting: Fixed.

src/main/python/ttconv/tt.py
palemieux: What about passing a reading/writing flag as a parameter to `LOGGER.info()` on line 116? This would allow reading and writing progress bars to be active simultaneously, e.g. if reading and writing were multithreaded.

src/test/python/test_progress_logging.py
palemieux: > That's pretty cool. It doesn't seem to work with my mucking around with the positioning of the progress bar.

Too bad. What does `cm.output` look like?

palemieux (approved):

src/main/python/ttconv/tt.py
```diff
@@ -93,38 +99,46 @@ def emit(self, record):
...
         self.is_writing_progress_bar = True
         if percent_progress is not None and float(percent_progress) >= 1.0:
-          sys.stdout.write(str('\r\n'))
+          sys.stdout.write('\r\n')
           self.is_writing_progress_bar = False
 
       self.flush()
 
     except (KeyboardInterrupt, SystemExit):
       raise
```
BobBrownConsulting, line 117: That's cool you can suppress that one line.

Issue #189, opened 2020-12-20
Layout and head elements emitted even if empty


palemieux:
See, for example, `build\imsc1\ttml\br\Br001.ttml` as emitted by the unit tests.

```
<ttml:tt xmlns:ttml="http://www.w3.org/ns/ttml" xml:lang="en">
  <ttml:head>
    <ttml:layout/>
  </ttml:head>
  <ttml:body end="00:00:10.000">
    <ttml:div end="00:00:10.000">
      <ttml:p end="00:00:10.000">
        <ttml:span>This text must be on the first line.</ttml:span>
        <ttml:br/>
        <ttml:span>This text on a second line.</ttml:span>
      </ttml:p>
    </ttml:div>
  </ttml:body>
</ttml:tt>
```

BobBrownConsulting:
I made a fix. 

New result:

<?xml version="1.0" encoding="UTF-8"?>
<ttml:tt xmlns:ttml="http://www.w3.org/ns/ttml" xml:lang="en">
   <ttml:body end="00:00:10.000">
      <ttml:div end="00:00:10.000">
         <ttml:p end="00:00:10.000">
            <ttml:span>This text must be on the first line.</ttml:span>
            <ttml:br />
            <ttml:span>This text on a second line.</ttml:span>
         </ttml:p>
      </ttml:div>
   </ttml:body>
</ttml:tt>

PR #202, opened 2020-12-21
Fixing writing of an empty head (#189)

BobBrownConsulting:


src/main/python/ttconv/imsc/elements.py
```diff
@@ -362,16 +362,20 @@ def from_model(
     `ctx` contains state information used in the process.
     '''
 
-    head_element = et.Element(HeadElement.qn)
+    head_element = None
 
     styling_element = StylingElement.from_model(ctx, model_doc)
 
     if styling_element is not None:
```
BobBrownConsulting, line 369: There might be a more python-y way to fix this. Feel free to improve or request changes. 
The root cause was creating the head  and layout elements when they were not needed.

palemieux: Looks good. I have made a small addition to catch the corner case where an empty `styling` element would be added if a child `initial` element was not output due to an error.

palemieux (approved):

PR #241, opened 2021-03-24
SCC reader sets ActiveArea to the safe area

### LINKED ISSUE #238: SCC reader should set ActiveArea to the safe area
https://www.w3.org/TR/ttml-imsc1.1/#ittp-activeArea

valnoel:
Closes #238

### DIFF
--- a/src/main/python/ttconv/scc/reader.py
+++ b/src/main/python/ttconv/scc/reader.py
@@ -30,7 +30,7 @@ from __future__ import annotations
 import logging
 from typing import Optional, List
 
-from ttconv.model import ContentDocument, Body, Div, CellResolutionType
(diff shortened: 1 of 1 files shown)


palemieux (changes requested):
> `ittp:activeArea="10% 10.5263% 80% 78.9474%"`

Why the active area is not exactly `"10% 10% 80% 80%"``?

valnoel:
> 
> 
> > `ittp:activeArea="10% 10.5263% 80% 78.9474%"`
> 
> Why the active area is not exactly `"10% 10% 80% 80%"``?

Because the SCC safe area is 32 x 15 cells over 40 x 19 cells, which gives 80% x 78.9474%. Is that a problem?

palemieux (approved):

PR #273, opened 2021-08-12
Misc STL fixes

### LINKED ISSUE #271: stl-data-processing: Set line height on p content element instead of body
Currently, the default line-height is set on the body content element:

https://github.com/sandflow/ttconv/blob/ca19bf25512d33cc4c35b5f5520edc6727cefb80/src/main/python/ttconv/stl/datafile.py#L113

This may lead to conflicts where a font size of a p content element is set to a value greater than 125%. As the line height is computed based on the elements font size where line-height is specified it would be in this case (if the font size of the body is not set explicitly) 125% of 100%. If the font size is later changed on a p element to a value greater than 125% (e.g. 160% for double height teletext rows) the line height will be smaller than the font size. In a multiline subtitle block, lines would overlap.

### LINKED ISSUE #272: stl-data-processing: Correct Typo in setting vertical alignment (exchange "after" and "before")
The alignment of the region is calculated based on the vertical position (vp) divided by half of the maximum number of Teletext rows (mnr). If a vertical position is above the middle of the teletext grid (vp < mnr / 2) the text is top aligned otherwise bottom aligned. But in the current implementation displayAlign in the top aligned region is set to `after` and in the bottom aligned region to `before`. To get the right alignment it should be the other way around.

https://github.com/sandflow/ttconv/blob/ca19bf25512d33cc4c35b5f5520edc6727cefb80/src/main/python/ttconv/stl/datafile.py#L204-L218

For the top aligned region it should be instead of `styles.DisplayAlignType.after`  ->  `styles.DisplayAlignType.before`.

palemieux:
Fix STL vertical alignment (closes #272)
Fix STL line height (closes #271)

### DIFF
--- a/src/main/python/ttconv/stl/datafile.py
+++ b/src/main/python/ttconv/stl/datafile.py
@@ -110,8 +110,6 @@ class DataFile:
 
     self.body = model.Body(self.doc)
     
-    self.body.set_style(styles.StyleProperties.LineHeight, styles.LengthType(DEFAULT_LINE_HEIGHT_PCT, styles.LengthType.Units.pct))
-
     self.doc.set_body(self.body)
 
     self.sgn_to_div_map = {}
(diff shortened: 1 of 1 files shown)


xchange11 (approved):
Tested and looks good to me! Thanks for the fix!

PR #278, opened 2021-08-12
Add tts:fillLineGap option to STL reader

### LINKED ISSUE #276: STL-Conversion: Add feature similiar to itts:fillLineGap
Consecutive lines in Teletext subtitles do not have blank space between them. Therefore a feature similar to [itts:fillLineGap](https://www.w3.org/TR/ttml-imsc1.1/#itts-fillLineGap) of the IMSC spec should be used when converting EBU STL to the ttconv model. 

It could be implemented as follows:
* specification on the body content element so that it is applied to all p content element
* value is by default true (always switched on)
* there should be a parameter to overwrite this because in rare cases multiple new lines are used between consecutive subtitle lines where blank space between the lines is intended.

palemieux:
Closes #276

### DIFF
--- a/src/test/python/test_stl_reader_config.py
+++ b/src/test/python/test_stl_reader_config.py
@@ -23,7 +23,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-'''Unit tests for the STL reader convfiguration'''
+'''Unit tests for the STL reader configuration'''
 
 # pylint: disable=R0201,C0115,C0116
 
--- a/src/test/python/test_stl_reader_config.py
+++ b/src/test/python/test_stl_reader_config.py
@@ -23,7 +23,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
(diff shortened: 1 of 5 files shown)


xchange11 (approved):
* Checked rendering and config option. Looks good to me. Thanks! 
* One thing to consider: latest when the feature branch is merged into the master the new config option should be added to the README.
* One small typo with a fix (does not prevent merging)

src/test/python/test_stl_reader_config.py
```diff
@@ -0,0 +1,62 @@
...
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+# ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+'''Unit tests for the STL reader convfiguration'''
```
xchange11, line 26:
```suggestion
'''Unit tests for the STL reader configuration'''
```

PR #282, opened 2021-08-13
Add support for line padding in STL reader

### LINKED ISSUE #277: STL Conversion: Adding line padding to each line
Because of the startbox and endbox control codes in Teletext an extra space with the background color of the line is added to each Teletext row. Intended or not: this behavior improves the readability of the subtitle. Therefore for TTML formats the feature  `ebutts:linePadding` was introduced. This feature should also be used when converting EBU STL to the ttconv model.

One strategy could be :
* apply line padding by default,
* specify it on the body element so it is inherited to all p elements,
* set width of half the space of a Teletext character (e.g. if a region represents the teletext grid -> region width / 40 / 2)
* the feature can be switched off (because line padding adds padding space to a line and could potentially lead to overflow of the region)

A more detailed discussion with advanced calculation strategies for line padding can be found in the [EBU-TT-D spec](https://tech.ebu.ch/docs/tech/tech3380.pdf) Annex D and in the [BBC Subtitle Guidelines](https://bbc.github.io/subtitle-guidelines/#ebutts-linePadding).

palemieux:
Closes #277


andreastai (approved):
Checked code, rendering, and use of config. Looks good to me. Thanks for the implementation!

xchange11 (approved):

PR #283, opened 2021-08-14
Override the TCP field of the GSI

### LINKED ISSUE #274: STL: subtitles earlier than TCP should be ignored
TTI blocks with a timecode earlier than _Time Code: Start-of-Programme (TCP)_ are used to store metadata and should be ignored by default.

A configuration parameter should be available to override this behavior.

palemieux:
Closes #274


palemieux:
@xchange11 Any chance you might have tests with two subtitles, one before the TCP and one after the TCP?

palemieux:
@xchange11 `test_irt_requirement_0061_004` and `test_irt_requirement_0062_002` now fail. It looks like the GSI has a non-zero TCP, but the tests do not take it into account when computing the expected begin/end times of the subtitles:

`self.assertEqual(p.get_end(),
                       SmpteTimeCode.parse("23:59:59:24",FPS_25).to_temporal_offset())`

xchange11 (approved):
Looks good to me!

PR #294, opened 2021-08-17
Add tests cumulative

### LINKED ISSUE #291: Find STL test files that use cumulative/add-on subtitles
The STL reader code implements supports but no unit tests are available

xchange11:
Closes #291


palemieux:
@xchange11 `cumulative_set.stl` might two issues:
* SN is repeated between the first and second TTI
* TNB field seems to have an incorrect value (1 instead of 5)

palemieux:
Replaced by https://github.com/sandflow/ttconv/pull/296

PR #296, opened 2021-08-17
Fix cumulative STL sub handling

### LINKED ISSUE #291: Find STL test files that use cumulative/add-on subtitles

### LINKED ISSUE #295: Initialize p element for cumulative subs and no br for last sub of cumulative set
https://github.com/sandflow/ttconv/blob/27846c9c8b1c40d5c787a833e07a33ab58c303dd/src/main/python/ttconv/stl/datafile.py#L303-L305
* The var `p_element` has not been declared and initialized at that point.
* The last subtitle of the cumulative set (with `0x03` in the `cs` field of the TTI block) should not have an appended `br` element.

palemieux:
Closes #295 
Closes #291


xchange11 (approved):
Nice! Teste and looks good to me!

xchange11 (commented):
After converting the model implementation to IMSC there is an issue with overlapping lines. This is caused by setting the font size on the span element but the line height to the parent p element.

PR #299, opened 2021-08-18
Fix fontSize for STL cumulative subtitles

### LINKED ISSUE #297: stl-reader: Overlapping lines with cumulative set
After converting the model implementation of the cumulative test resource to IMSC there is an issue with overlapping lines. This is caused by setting the font size on the span element but the line height to the parent p element.

palemieux:
Closes #297

### DIFF
--- a/src/main/python/ttconv/stl/datafile.py
+++ b/src/main/python/ttconv/stl/datafile.py
@@ -245,6 +245,18 @@ class DataFile:
         styles.LengthType.Units.pct)
       )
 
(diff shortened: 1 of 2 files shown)


xchange11 (changes requested):
For each cumulative row the line breaks/br content elements need also be timed accordingly. Currently, they are not part of the respective timed span element and they are added before the text content appears. See below a picture of the IMSC rendering based on the generated ttconv model.

image

xchange11 (approved):
Looks good to me. Thanks!

PR #304, opened 2021-08-18
Fix STL region height with double height characters

### LINKED ISSUE #300: stl-reader: Vertical positioning of multiple row subtitle with double height
A TTI Block with VP 18 and three lines/rows should be aligned at the bottom of the Teletext grid (the last line should be on Teletext rows 22 and 23). For the ttconv model the subtitle should be bottom aligned in a region of height 80% with 10%  margin at the top.

palemieux:
Closes #300


xchange11 (approved):
Thanks! Looks good to me!

PR #305, opened 2021-08-18
Protect progress bar against values outside of [0,1]

### LINKED ISSUE #302: Multiple progress bars when TNB field in the GSI block has incorrect values
The progress bar appears multiple times when the value for the GSI field Total Number of TTI Blocks (TNB) is less than the actual number of TTI blocks in the STL file (e.g. when TNB is 1 and the number of TTI blocks is 11 it appears eleven times, see picture below). Although it is an error in the STL file it would be good that ttconv is robust against these errors as values in the GSI block are not always reliable.

image

palemieux:
Closes #302


xchange11 (approved):
Applied the change to the STL feature branch and tested against file with incorrect TNB values. Looks good.

PR #306, opened 2021-08-18
Ignore TCP GSI field by default

### LINKED ISSUE #301: Ignore TCP by default
Incorrect TCP values, e.g., `10:00:00:00`, result in empty output.

palemieux:
Closes #301


xchange11 (approved):
Tested with several configuration data for `program_start_tc` and works fine. Thanks.

PR #310, opened 2021-08-19
Fix STL reader language support

### LINKED ISSUE #308: LC field is a two-character ASCII string and xml:lang is never set on root element
https://github.com/sandflow/ttconv/blob/3115b7a781177be896f2190f2298ebbd906bf71c/src/main/python/ttconv/stl/datafile.py#L322

should be `int(self.gsi.LC, 16)`

palemieux:
Closes #308


xchange11 (approved):
Tested with several values for the `LC` field and looks good to me.

PR #319, opened 2021-08-20
Fix reset of styling after newline when reading STL files

palemieux:


xchange11 (changes requested):
Background and text color needs to be reset after a new line. See also the corresponding tests in PR #321

PR #321, opened 2021-08-23
Add tests for style reset after newline

xchange11:
- After a new line the color coding is reset for background and
  text color.


palemieux (changes requested):
Could we the unit test use styles/colors that are unlikely to be default, e.g. red for foreground and green for background and italic for font style?

palemieux:
> A new test could be added where the colors of the second row are set explicitly and differ from the first row (done with b0cf0b6).

Perfect.

Ideally we would have a third test with two lines, both the same combination of  (non-default) italic/fg color/bg colors.

PR #328, opened 2021-09-03
SCC: Use the CEA-608 standard characters mapping instead of ASCII table

### LINKED ISSUE #326: CEA 608 "FULL BLOCK" mapping missing?
Is the mapping to "FULL BLOCK" character missing from the SCC reader?

image

https://github.com/sandflow/ttconv/blob/master/src/main/python/ttconv/scc/codes/special_characters.py

valnoel:
Closes #326

### DIFF
--- a/src/test/python/test_scc_standard_characters.py
+++ b/src/test/python/test_scc_standard_characters.py
@@ -32,9 +32,9 @@ import unittest
 from ttconv.scc.codes.standard_characters import SCC_STANDARD_CHARACTERS_MAPPING
 
 
-class SccExtendedCharacterTest(unittest.TestCase):
+class SccStandardCharactersTest(unittest.TestCase):
 
-  def test_scc_special_character_values(self):
+  def test_scc_standard_character_values(self):
     self.assertEqual(SCC_STANDARD_CHARACTERS_MAPPING[0x20], " ")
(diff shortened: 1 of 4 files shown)


src/test/python/test_scc_standard_characters.py
```diff
@@ -0,0 +1,133 @@
...
+"""Unit tests for the SCC Standard characters mapping"""
+
+# pylint: disable=R0201,C0115,C0116,W0212
+
+import unittest
+
+from ttconv.scc.codes.standard_characters import SCC_STANDARD_CHARACTERS_MAPPING
+
+
+class SccExtendedCharacterTest(unittest.TestCase):
```
palemieux, line 35:
```suggestion
class SccStandardCharacterTest(unittest.TestCase):
```

src/test/python/test_scc_standard_characters.py
```diff
@@ -0,0 +1,133 @@
...
+
+  def test_scc_special_character_values(self):
```
palemieux, line 37:
```suggestion
  def test_scc_standard_character_values(self):
```

palemieux (changes requested):

palemieux (approved):

PR #352, opened 2021-11-29
TTML writer: support time-clock-with-frames output

### LINKED ISSUE #351: TTML writer should output in timecode hh:mm:ss:ff unit
At this time it seems not possible to output the TTML subtitle timestamps in a timecode format with frames unit (and not ms), based on the EBU STL source frame rate. A such enhancement should be welcome for using ttconv within the broadcast workflows.

palemieux:
Closes #351

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,27 +62,69 @@ tt convert -i <input .scc file> -o <output .ttml file>
 
 ### Command line
 
-```
-tt convert [-h] -i INPUT -o OUTPUT [--itype ITYPE] [--otype OTYPE] [--config CONFIG] [--config_file CONFIG_FILE]
-```
+`tt convert [-h] -i INPUT -o OUTPUT [--itype ITYPE] [--otype OTYPE] [--config CONFIG] [--config_file CONFIG_FILE]`
 
 * `--itype`: `TTML` | `SCC` | `STL` | `SRT` (extrapolated from the filename, if omitted)
 * `--otype`: `TTML` | `SRT` | `VTT` (extrapolated from the filename, if omitted)
 * `--config` and `--config_file`: JSON dictionaries with the following members:
-
   * `"general"."progress_bar": "true" | "false"`: whether a progress bar is displayed
   * `"general"."log_level": "INFO" | "WARN" | "ERROR"`: logging level
-  * `"imsc_writer"."time_format": "frames" | "clock_time" | "clock_time_with_frames" (default: "clock_time")`: output TTML expressions in frames (`f`), as `HH:MM:SS.mmm` or as `HH:MM:SS:FF`
-  * `"imsc_writer"."fps": "<num>/<denom>"`: specifies the frame rate _num/denom_ when `time_format` is `frames` or `clock_time_with_frames`
-  * `"stl_reader"."disable_fill_line_gap" : "true" | "false" (default: "false")`: "true" means that the STL reader does not fill gaps between lines
-  * `"stl_reader"."disable_line_padding" : "true" | "false" (default: "false")`: "true" means that the STL reader does not add padding at the begining/end of lines
-  * `"stl_reader"."program_start_tc" : "TCP" | "HH:MM:SS:FF" (default: "00:00:00:00")`: specifies a starting offset, either the TCP field of the GSI block or a user-specified timecode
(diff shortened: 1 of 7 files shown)


xchange11 (commented):
Implementation looks good to me.

Some proposals:

* the `fps` parameter is mandatory when using `time_format` = `clock_time_with_frames` or `frames`.
  * This should be added as a requirement to the documentation.
  * If `fps` is missing I get the error message `AttributeError: 'NoneType' object has no attribute 'split'`. This should be an error message that points to the missing `time_format` parameter.
* The `time_format` parameter is mandatory when `fps` is specified. 
  * The `time_format` parameter should be optional. When the default of `time_format` is applied (`clock_time`), then the `fps` should just be ignored. This is already the case when `time_format` is defined explicitly together with `fps`.
* Some examples of how to use `time_format` and `fps` together may help.

Vincent-Dabouineau (approved):
Perfect and tested, works like a charm!

Vincent-Dabouineau (approved):

palemieux:
> the fps parameter is mandatory when using time_format = clock_time_with_frames or frames.
> This should be added as a requirement to the documentation.
> If fps is missing I get the error message AttributeError: 'NoneType' object has no attribute 'split'. This should be an error message that points to the missing time_format parameter.

+1

> The time_format parameter is mandatory when fps is specified.

I think this would break backward compatibility where the presence for `fps` implied `frames`.

> Some examples of how to use time_format and fps together may help.

+1

Vincent-Dabouineau:
> > the fps parameter is mandatory when using time_format = clock_time_with_frames or frames.
> > This should be added as a requirement to the documentation.
> > If fps is missing I get the error message AttributeError: 'NoneType' object has no attribute 'split'. This should be an error message that points to the missing time_format parameter.
> 
> +1
> 
> > The time_format parameter is mandatory when fps is specified.
> 
> I think this would break backward compatibility where the presence for `fps` implied `frames`.
> 
> > Some examples of how to use time_format and fps together may help.
> 
> +1

I always use the `fps`parameter so it works. By the way, I guess the frame rate is encapsulated in EBU STL sources, so it should be retrieved/used automatically?

palemieux:
> I always use the fps parameter so it works. By the way, I guess the frame rate is encapsulated in EBU STL sources, so it should be retrieved/used automatically?

The internal model does not preserve frame rate, so you can set it to whatever you want on output.

palemieux:
@xchange11 Please review revised documentation and error reporting.

README.md
```diff
@@ -62,28 +62,70 @@ tt convert -i <input .scc file> -o <output .ttml file>
...
-```
-tt convert [-h] -i INPUT -o OUTPUT [--itype ITYPE] [--otype OTYPE] [--config CONFIG] [--config_file CONFIG_FILE]
-```
+`tt convert [-h] -i INPUT -o OUTPUT [--itype ITYPE] [--otype OTYPE] [--config CONFIG] [--config_file CONFIG_FILE]`
 
 * `--itype`: `TTML` | `SCC` | `STL` | `SRT` (extrapolated from the filename, if omitted)
 * `--otype`: `TTML` | `SRT` | `VTT` (extrapolated from the filename, if omitted)
 * `--config` and `--config_file`: JSON dictionaries with the following members:
-
   * `"general"."progress_bar": "true" | "false"`: whether a progress bar is displayed
```
xchange11, line 70: You may want to put the general configuration under a separate heading like you did for `imsc_writer` and  `stl_reader`.

README.md
```diff
@@ -62,28 +62,70 @@ tt convert -i <input .scc file> -o <output .ttml file>
...
 Example:
 
 `tt convert -i <.scc file> -o <.ttml file> --itype SCC --otype TTML --config '{"general": {"progress_bar":false, "log_level":"WARN"}}'`
 
+### IMSC Writer configuration
+
+* `"time_format": "frames" | "clock_time" | "clock_time_with_frames"` \[optional\]
+
+Specifies whether the TTML time expressions are in frames (`f`), `HH:MM:SS.mmm` or `HH:MM:SS:FF`
+
```
xchange11, line 84: For readability, it may be better to use nested list items e.g. 
* `"time_format": "frames" | "clock_time" | "clock_time_with_frames"` \[optional\]
  * Specifies whether the TTML time expressions are in frames (`f`), `HH:MM:SS.mmm` or `HH:MM:SS:FF`

PR #357, opened 2021-12-10
imsc writer: fix config parsing

### LINKED ISSUE #355: IMSC writer config: Error when timeformat is clock_time
This is due to an unhandled call for a `None` value in the  `FractionDecoder` 

https://github.com/sandflow/ttconv/blob/a87152c91647999e58606508beeecc633ad5448a/src/main/python/ttconv/imsc/config.py#L55-L58

The following fixes the issue:

```python
    def __call__(self, value: str) -> Fraction:
      if value is None:
        return None
      [num, den] = value.split('/')

      return Fraction(int(num), int(den))
```

This is also related to https://github.com/sandflow/ttconv/pull/352#pullrequestreview-819848021. The error handling improvement of 4bf4239 will never be reached because of the above error handling. After the fix 4bf4239 works fine.

palemieux:
Closes #355

### DIFF
--- a/src/main/python/ttconv/imsc/config.py
+++ b/src/main/python/ttconv/imsc/config.py
@@ -52,7 +52,10 @@ class IMSCWriterConfiguration(ModuleConfiguration):
   class FractionDecoder:
     """Utility callable for converting string to Fraction"""
 
-    def __call__(self, value: str) -> Fraction:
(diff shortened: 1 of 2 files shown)


xchange11 (approved):
Looks good to me.

PR #361, opened 2022-01-19
tt: set default log_level to "INFO"

### LINKED ISSUE #281: Progress-bar activation with config option
Progress-bar can not be switched on with config option e.g. 

```
 tt.py convert -i  [...]  -o [...] --config '{"general": {"progress_bar": true}}'
```

There seem to be some side effects with the logging option, because this works

```
 tt.py convert -i  [...]  -o [...]  '{"general": {"progress_bar": true, "log_level": "INFO"}}'
```

If the intention is to show the progress bar only with `"log_level": "INFO"` this should be documented in the README.

palemieux:
Closes #281

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Example:
 
 `"progress_bar": true | false`
 
-Specifies whether a progress bar is displayed
+A progress bar is displayed if `progress_bar` is `true` and `log_level` is `"INFO"`.
 
 Default: `true`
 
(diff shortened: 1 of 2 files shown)


xchange11 (commented):
What is the intended behaviour of the display of the progress bar? It is only displayed when the log level is "INFO" but not with log levels "WARN" and "ERROR". If this is the intended behaviour it may be possible to delete this parameter entirely. In this case, it will be always switched on in "INFO" mode. If the parameter is kept it should be documented that it only has an effect when the log level is "INFO".

palemieux:
@xchange11 Done at https://github.com/sandflow/ttconv/pull/361/commits/7f6308f777c2318d91c191db5f13d9ca020fb454

xchange11 (approved):
Thanks for the change! Looks good to me!

Issue #366, opened 2022-03-31
Request to remove 'ttml' prefix from converted TTML/IMSC documents


jbaden:
Hello, 

I am using TTCONV's SCC to TTML CLs and noticed that all content elements within a converted IMSC document include the following 'ttml' prefix which is causing validation errors within our tools. Would it be possible to remove these in favor of the general vocabulary documented in the [TTML2 spec](https://www.w3.org/TR/2018/REC-ttml1-20181108/#content-element-vocabulary)? 
```
<ttml:tt xmlns:ebutts="urn:ebu:tt:style"
    xmlns:ittp="http://www.w3.org/ns/ttml/profile/imsc1#parameter"
    xmlns:ttml="http://www.w3.org/ns/ttml" xmlns:ttp="http://www.w3.org/ns/ttml#parameter"
    xmlns:tts="http://www.w3.org/ns/ttml#styling" xml:lang="en" ttp:cellResolution="40 19"
    ittp:activeArea="10% 10.5263% 80% 78.9474%">
    <ttml:head>
        <ttml:layout>
            <ttml:region xml:id="pop1" tts:displayAlign="before" tts:extent="40% 5%"
                tts:origin="30% 84%" tts:showBackground="whenActive"/>
        </ttml:layout>
    </ttml:head>
    <ttml:body tts:fontFamily="&quot;Consolas&quot;, &quot;Monaco&quot;, monospace"
        tts:lineHeight="100%" ebutts:linePadding="0.25c">
        <ttml:div>
            <ttml:p region="pop1" begin="00:00:01.200" end="00:00:03.533" xml:id="caption1"
                tts:textAlign="start">
                <ttml:span tts:backgroundColor="#000000">(clanging metal)</ttml:span>
            </ttml:p>
```

PR #368, opened 2022-03-31
TTML writer: use the tt ns as the default ns (#366)

palemieux:


palemieux:
@jbaden For your review

jbaden (approved):

cconcolato (approved):

PR #383, opened 2023-01-26
Check for buffered_caption being None in SCC reader

atsampson:
This fixes a problem found when parsing CC data from a noisy LaserDisc: if a signal dropout corrupts the preamble for a caption, the SCC reader shouldn't crash when text is seen for that caption.

### DIFF
--- a/src/main/python/ttconv/scc/reader.py
+++ b/src/main/python/ttconv/scc/reader.py
@@ -455,7 +455,7 @@ class _SccContext:
       self.active_caption.get_current_text().add_style_property(StyleProperties.FontStyle, self.current_font_style)
       self.active_caption.get_current_text().add_style_property(StyleProperties.TextDecoration, self.current_text_decoration)
 
-    else:
+    elif self.buffered_caption is not None:
       self.buffered_caption.append_text(word)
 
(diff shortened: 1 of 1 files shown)


palemieux:
It looks like the code throws in other places when no caption is initialized:

* https://github.com/sandflow/ttconv/blob/0da12426c8e56b3079cc23fa721e23d7cb2357ad/src/main/python/ttconv/scc/reader.py#L296
* https://github.com/sandflow/ttconv/blob/0da12426c8e56b3079cc23fa721e23d7cb2357ad/src/main/python/ttconv/scc/reader.py#L244

@atsampson Is your proposal to basically ignore 608/708 codes until the next caption?

@valnoel Any opinion?

valnoel:
Hi @atsampson @palemieux !

I've completed the @atsampson suggestion by skipping captions parsing in case of no PAC is detected from the SCC stream (i.e. when the cursor position is not specified). I've added a unit test as well.

See https://github.com/sandflow/ttconv/pull/385

GRDazzle:
I have run into this problem as well. I can confirm the PR change fix the issue for me.

GRDazzle (approved):
This is tested and verified with my own test case.

palemieux:
Mooted by https://github.com/sandflow/ttconv/pull/395

PR #385, opened 2023-02-17
Fix: SCC reader skips captions without Preamble Address Code

valnoel:


palemieux:
Mooted by https://github.com/sandflow/ttconv/pull/395

PR #392, opened 2023-07-27
SCC: support word repetition for normal and extended characters

### LINKED ISSUE #393: SCC: support word repetition for normal and extended characters
I notice that some characters are skipped due to SCC_word repetition.

See https://github.com/sandflow/ttconv/pull/392#issue-1824674885

humberthardy:
I notice that some characters are skipped due to SCC_word repetition. 

Here are some examples: 
- SCC word "em": `remember` is (sometimes) translated in `rember`
- SCC word "ti": `repetition` can be translated in `repetion`
...

PS: My knowledge about SCC is low, I might miss something :)

Closes #393


palemieux:
Thanks for the report. Will process shortly. I have opened dedicated issue #393 .

src/main/python/ttconv/scc/reader.py
```diff
@@ -481,7 +481,8 @@ def process_line(self, line: SccLine) -> SmpteTimeCode:
 
     for scc_word in line.scc_words:
 
-      if self.previous_code == scc_word.value:
+      if self.previous_code == scc_word.value and self.previous_code_type is not str:
```
palemieux, line 484: I am thinking we should add a `isControlCode()` in `SccWord`, which would test whether the MSB is a control character.

palemieux (changes requested):

src/main/python/ttconv/scc/reader.py
palemieux: @humberthardy Should I propose an update to your branch?

palemieux (approved):

valnoel (changes requested):
A blank line missing at the end of the word.py file.

Otherwise, it looks OK to me!

PR #395, opened 2023-08-16
SCC: caption management refactoring

valnoel:
SCC reader refactoring proposal, addressing https://github.com/sandflow/ttconv/pull/383 & https://github.com/sandflow/ttconv/pull/385 issues


src/main/python/ttconv/scc/context.py
```diff
@@ -0,0 +1,479 @@
...
+    # Previously read SCC word value
+    self.previous_word: Optional[SccWord] = None
+    self.previous_word_type: Optional[Type] = None
+
+    # Buffered caption being built
+    self.buffered_caption = None
+    # Captions being displayed
+    self.active_caption: Optional[SccCaptionParagraph] = None
+    # Caption style (Pop-on, Roll-up, Paint-on) currently processed
+    self.current_style = SccCaptionStyle.Unknown
```
palemieux, line 71: Shouldn't we have a "default" style?

src/main/python/ttconv/scc/context.py
```diff
@@ -0,0 +1,479 @@
...
+    # Screen safe area offsets
+    self.safe_area_x_offset = safe_area_x_offset
+    self.safe_area_y_offset = safe_area_y_offset
+
+    # Previously read SCC word value
+    self.previous_word: Optional[SccWord] = None
+    self.previous_word_type: Optional[Type] = None
+
+    # Buffered caption being built
+    self.buffered_caption = None
```
palemieux, line 67: Isn't that redundant with `new_buffered_caption()` below?

src/main/python/ttconv/scc/context.py
```diff
@@ -0,0 +1,479 @@
...
+    # Captions being displayed
+    self.active_caption: Optional[SccCaptionParagraph] = None
```
palemieux, line 69: Shouldn't we start with a non-empty active caption, esp. if the default style is paint-on?

palemieux (changes requested):

src/main/python/ttconv/scc/context.py
valnoel: What would be this default style? It seems a bit tricky to choose...

palemieux: I skimmed the 608 spec and couldn't find an obvious answer.

I would choose paint-on for now.

src/main/python/ttconv/scc/context.py
palemieux: Shouldn't the begin time code be set when the first printable character is displayed, instead of when the caption is instantiated?

palemieux: The idea is to make the start of processing the same as the middle, i.e. there is always a front buffer and a back buffer active.

Maybe this is too much of a change for now. Let me know.

valnoel: > Shouldn't the begin time code be set when the first printable character is displayed, instead of when the caption is instantiated?

This is done for Paint-On only. The begin time must be set on EOC for Pop-On, and on CR for Roll-Up.

> The idea is to make the start of processing the same as the middle, i.e. there is always a front buffer and a back buffer active.

I understand the idea, but I think we cannot start writing into buffers without a minimum of context (a style at least), otherwise we're moving away from the spirit of the 608 specification...

But if it's a prerequisite, the change shouldn't be too big...

palemieux: 608/708 are in fact streaming format (SCC added the timestamps) and a device could join the stream at any time and thus possibly in the middle of a caption.

What is the current behavior if we start parsing in the middle of a caption? Maybe we should add this to the unit tests?

valnoel: @palemieux I've changed the reader behavior so that it skips captions until a style is specified in 49d14db

Let me know what you think!

palemieux:
@atsampson @rktcc Your feedback would be welcome on this PR.

palemieux:
@atsampson @rktcc Did you have the opportunity to review? The plan is to merge this PR sometime next week.

palemieux (approved):

palemieux (changes requested):
`itt.scc` looks incorrect when converted to TTML

palemieux (approved):

palemieux (approved):

PR #396, opened 2023-08-21
SCC: Use quotation mark instead of 2 backquote chars

humberthardy:
Hello,
I notice that SCC character "quotation mark" is translated into two backquote characters ` `` ` instead of a single quotation mark ` " `
 ( hex char `0x22` in https://en.wikipedia.org/wiki/EIA-608)
Is it a mistake?


palemieux (approved):

palemieux:
Looks like a bug in the CTA 608 spec:

https://github.com/sandflow/ttconv/pull/397

PR #402, opened 2023-09-12
Fixed wrong arg order in IMSC attr parsing for cell resolution

### LINKED ISSUE #403: IMSC reader: ttp:cellResolution components are swapped
This commit fixes the switched order of columns and rows in the IMSC attribute parsing for cellResolution.

Closes https://github.com/sandflow/ttconv/pull/402

chrsan:
This commit fixes the switched order of `columns` and `rows` in the IMSC attribute parsing for `cellResolution`.

Closes #403

### DIFF
--- a/src/main/python/ttconv/imsc/attributes.py
+++ b/src/main/python/ttconv/imsc/attributes.py
@@ -128,7 +128,7 @@ class CellResolutionAttribute:
 
       if m is not None:
 
-        return model.CellResolutionType(int(m.group(1)), int(m.group(2)))
+        return model.CellResolutionType(columns=int(m.group(1)), rows=int(m.group(2)))
 
       LOGGER.error("ttp:cellResolution invalid syntax")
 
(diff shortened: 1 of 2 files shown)


palemieux:
Can you add a unit test:

```
  def test_cell_resolution(self):
    xml_str = """<?xml version="1.0" encoding="UTF-8"?>
    <tt xml:lang="en"
        xmlns="http://www.w3.org/ns/ttml"
        xmlns:ttp="http://www.w3.org/ns/ttml#parameter"
        ttp:cellResolution="32 15">
    </tt>"""
    doc = imsc_reader.to_model(et.ElementTree(et.fromstring(xml_str)))
    self.assertEqual(doc.get_cell_resolution().columns, 32)
    self.assertEqual(doc.get_cell_resolution().rows, 15)
```

chrsan:
Thanks for the test case! I've added it in my latest commit.

chrsan:
@palemieux  BTW, I think the initial value for `ebutts:linePadding` is wrong. It should be `0c` and not `1c` as it is right now if I'm correct?

palemieux (approved):

PR #409, opened 2023-11-07
Misc

ngaullier:
Use case: output SDH TTML (basic, not EBU) and WebVTT from an EBU STL input.

Add features:
- STL reader : disable EBU style, force position to bottom
- TTML : fix missing xml declaration
- WebVTT : add support for 'align'
- misc missing documentation

### DIFF
--- a/src/main/python/ttconv/tt.py
+++ b/src/main/python/ttconv/tt.py
@@ -376,7 +376,7 @@ def convert(args):
     #
     # Write out the converted file
     #
-    tree_from_model.write(outputfile, encoding="utf-8")
+    tree_from_model.write(outputfile, encoding="utf-8", xml_declaration=True)
 
   elif writer_type is FileTypes.SRT:
     #
--- a/README.md
+++ b/README.md
@@ -209,6 +209,14 @@ Default: `true`
 
 Default: `false`
 
+#### text_position
+
+`"text_position" : true | false`
+
+`true` means that the VTT writer outputs text alignment cue settings
+
+Default: `false`
+
 #### cue_id
 
 `"cue_id" : true | false`
--- a/README.md
+++ b/README.md
@@ -171,6 +171,14 @@ Specifies a maximum number of rows for open subtitles, either the MNR field of t
 
 Default: `23`
 
+#### force_bottom
+
+`"force_bottom" : true | false`
+
+Overrides line positions, force to bottom
+
+Default: `false`
+
 ### SRT Writer configuration
 
 #### text_formatting
--- a/README.md
+++ b/README.md
@@ -147,6 +147,16 @@ Default: `false`
 
 Default: `false`
 
+#### disable_ebu_style
+
+`"disable_ebu_style" : true | false`
+
+`true` means that the STL reader does not try to map EBU style as described in EBU Tech 3360
+
+Implies `disable_line_padding`
+
+Default: `false`
+
 #### program_start_tc
 
 `"program_start_tc" : "TCP" | "HH:MM:SS:FF"`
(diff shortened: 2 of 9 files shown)


src/main/python/ttconv/tt.py
```diff
@@ -376,7 +376,7 @@ def convert(args):
     #
     # Write out the converted file
     #
-    tree_from_model.write(outputfile, encoding="utf-8")
```
palemieux, line 379: The XML declaration is optional in XML 1.0, which is the [default](https://www.w3.org/TR/ttml2/#concrete-encoding) for IMSC and TTML.

palemieux:
@ngaullier Can you separate each of the features in a separate pull request?

src/main/python/ttconv/tt.py
ngaullier: Well, I really don't know about the standards, I check the ttml with https://github.com/skynav/ttt
The result is that, without the xml declaration, I get encoding errors on some diacritical marks and french symbols (ç ...).
Is utf-8 the default when no xml declaration exists -> do you think it is a ttt issue which fails to defaults to utf-8 ? or should the particular encoding characters be detected to trigger the xml declaration ?

ngaullier:
Just to confirm before to proceed: you mean exactly 4 pull requests for each my bullet points above ?

src/main/python/ttconv/tt.py
palemieux: @ngaullier Can you provide a sample file that fails TTV validation? Looking at code I wrote that uses TTV, I force UTF-8:

```
      args.add("--force-encoding");
      args.add("UTF-8");
```

There is also a known issue with BOM (https://github.com/skynav/ttt/issues/193) but it is probably not relevant here.

palemieux:
> Just to confirm before to proceed: you mean exactly 4 pull requests for each my bullet points above ?

Yes, I suggest we start with _WebVTT : add support for 'align'_

src/main/python/ttconv/tt.py
ngaullier: [samplefrench.zip](https://github.com/sandflow/ttconv/files/13294150/samplefrench.zip)
PYTHONPATH=src/main/python python3 src/main/python/ttconv/tt.py convert -i /mnt/d/samplefrench.stl -o /mnt/d/samplefrench.ttml
>Timed Text Verifier (TTV) [7.2-SNAPSHOT] Copyright (c) 2013-21 Skynav, Inc.
[E]:Malformed US-ASCII at byte offset 815 of one byte.

On "é" of métisse:
10:00:45:06-10:00:47:07    SGN.SN.EBN.CS.VP=00.0012.FF.00.20C
   	[DOUBLE HEIGHT][ALPHA CYAN]          <<qui est métisse,>>|| (30)

Yes, I also experienced ttv/BOM issue, but personnaly I don't like BOMs anyway!

ngaullier:
Ok, so one at a time as I read you. Anyway the ttml is currently questionnable so maybe not a good idea to start a new PR for this one at the moment.

palemieux:
> samplefrench.zip

I am pretty sure that TTV defaults to US-ASCII.

ngaullier:
Understood, utf-8 should be the default. I will submit a new issue or PR to ttv.
I let you review the vtt at first, then I will submit the PR for STL.
Concerning the slight README updates, maybe it is better if you can handle/commit it yourself as a PR seems overkill and you may want to arrange the wording differently.
So now I think we can close this "misc" issue. Thanks for your reviews.

palemieux:
> Concerning the slight README updates, maybe it is better if you can handle/commit it yourself as a PR seems overkill and you may want to arrange the wording differently.
> So now I think we can close this "misc" issue. Thanks for your reviews.

+1

PR #410, opened 2023-11-08
WebVTT writer: add support for 'align' (#407)

### LINKED ISSUE #407: VTT: add support for TextAlign
My use case is "STL in to VTT out".
The STL parser seems to implement JC -> TextAlignType, but the TextAlignType is not taken into account by the VTT writer...
Thank you!
Nicolas

ngaullier:
Closes #407

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,14 @@ Default: `true`
 
 Default: `false`
 
+#### text_align
+
+`"text_align" : true | false`
+
+`true` means that the VTT writer outputs text alignment cue settings
+
+Default: `false`
+
 #### cue_id
 
 `"cue_id" : true | false`
--- a/src/main/python/ttconv/vtt/config.py
+++ b/src/main/python/ttconv/vtt/config.py
@@ -41,5 +41,8 @@ class VTTWriterConfiguration(ModuleConfiguration):
   # outputs `line` and `line alignment` cue settings
   line_position: bool = field(default=False, metadata={"decoder": bool})
 
+  # outputs `text alignment` cue settings
+  text_align: bool = field(default=False, metadata={"decoder": bool})
+
   # outputs cue identifier
   cue_id: bool = field(default=True, metadata={"decoder": bool})
--- a/src/main/python/ttconv/vtt/cue.py
+++ b/src/main/python/ttconv/vtt/cue.py
@@ -42,6 +42,12 @@ class VttCue:
     center = "center"
     end = "end"
 
+  class TextAlignment(Enum):
+    """WebVTT text alignment cue setting"""
+    left = "left"
+    center = "center"
+    right = "right"
+
   _EOL_SEQ_RE = re.compile(r"\n{2,}")
 
   def __init__(self, identifier: Optional[int] = None):
@@ -50,7 +56,8 @@ class VttCue:
     self._end: Optional[ClockTime] = None
     self._text: str = ""
     self._line: int = None
-    self._align: VttCue.LineAlignment = None
+    self._linealign: VttCue.LineAlignment = None
+    self._textalign: VttCue.TextAlignment = None
 
   def set_begin(self, offset: Fraction):
     """Sets the paragraph begin time code"""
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -191,6 +191,14 @@ Default: `true`
 
 Default: `false`
 
+#### text_position
+
+`"text_position" : true | false`
```
palemieux, line 196:
```suggestion
`"text_align" : true | false`
```

src/main/python/ttconv/vtt/cue.py
```diff
@@ -42,6 +42,12 @@ class LineAlignment(Enum):
     center = "center"
     end = "end"
 
+  class TextAlignment(Enum):
+    """WebVTT text alignment cue setting"""
```
palemieux, line 46: That should be  "start", "center", "end", "left", "right"

https://www.w3.org/TR/webvtt1/#webvtt-alignment-cue-setting

src/test/python/test_vtt_writer.py
```diff
@@ -114,21 +114,25 @@ def test_position(self):
   <head>
     <styling>
       <style xml:id="style.center" tts:fontFamily="Arial" tts:fontSize="100%" tts:fontStyle="normal" tts:fontWeight="normal" tts:backgroundColor="transparent" tts:color="white" tts:textAlign="center"/>
+      <style xml:id="style.left" tts:fontFamily="Arial" tts:fontSize="100%" tts:fontStyle="normal" tts:fontWeight="normal" tts:backgroundColor="transparent" tts:color="white" tts:textAlign="left"/>
```
palemieux, line 117: I think this should be a different test, since this one is for position.

palemieux (changes requested):

src/main/python/ttconv/vtt/config.py
```diff
@@ -41,5 +41,8 @@ def name(cls):
   # outputs `line` and `line alignment` cue settings
   line_position: bool = field(default=False, metadata={"decoder": bool})
 
+  # outputs `text alignment` cue settings
+  text_position: bool = field(default=False, metadata={"decoder": bool})
```
palemieux, line 45:
```suggestion
  text_align: bool = field(default=False, metadata={"decoder": bool})
```

src/main/python/ttconv/vtt/cue.py
```diff
@@ -51,6 +57,7 @@ def __init__(self, identifier: Optional[int] = None):
     self._text: str = ""
     self._line: int = None
     self._align: VttCue.LineAlignment = None
```
palemieux, line 59:
```suggestion
    self._line_align: VttCue.LineAlignment = None
```

src/main/python/ttconv/vtt/cue.py
ngaullier: Okay, "middle" does not exist... I took a production sample as reference, I should not have!
In the updated PR, I kept the "left/right" notation because it seems more commonpractice with vtt (which inherited from srt with similar "pad" \anX tricks), despite modern TTML "prefer" start/end, I feel.
BTW, writing the test code, I realized that current TTML reader code parse "left/right" to "start/end" without setting writing mode which would be an issue for rl; but I think the TTML case "left/right" with right-left reading mode does not match many real world use cases.

ngaullier:
I think I have applied all of your suggestions. Thank you for the review.

src/main/python/ttconv/vtt/writer.py
```diff
@@ -175,6 +181,16 @@ def process_p(self, region: ISD.Region, element: model.P, begin: Fraction, end:
         cue.set_line(round(position.v_offset.value + extent.height.value / 2))
         cue.set_align(VttCue.LineAlignment.center)
 
+    if self._config.text_align:
```
palemieux, line 184: I think you also need to take into consideration `tts:direction` if you want to resolve `start` and `end`:

https://github.com/sandflow/imscJS/blob/32d0767ba0830dbac984ff9584c45934b0b8cd45/src/main/js/html.js#L270

palemieux (changes requested):

ngaullier:
Well, I am not comfortable with right-left tricks. The writing_mode can conflict with direction and both have default initial values, so it is not crystal clear to me how this should be handled. I check the result with your ownline tool
https://www.sandflow.com/imsc1proc/index.html
So hopefully I get it ?
I have added an exception where a direction is specifically defined on a span, as it would require a complete rewrite of some blocks of code and seems overly complicated for what I see as an exotic unlikely scenario.

palemieux:
@ngaullier I am reviewing the PR and think I have found a bug in ttconv. Will report shortly.

palemieux:
@ngaullier Check https://github.com/sandflow/ttconv/pull/414 please

ngaullier:
Forwarded to PR #414

PR #414, opened 2023-11-17
WebVTT writer: add support for 'align' (proposed fix)

palemieux:
Closes #410

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,14 @@ Default: `true`
 
 Default: `false`
 
+#### text_align
+
+`"text_align" : true | false`
+
+`true` means that the VTT writer outputs text alignment cue settings
+
+Default: `false`
+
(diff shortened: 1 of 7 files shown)


ngaullier:
Great, thank you!
In fact, I realized that I... forgot to push my changes to the PR #410... But anyway, it was not very good looking to implement both writingmode and direction in the VTT code as I implemented it.
Now, I understand how the two interacts, I have learned one thing!
Oviously, the code is much better and easier understandable this way.
I go close my original PR.
Only thing that should be kept in mind is that span-direction is not supported (I think it "cannot" be supported in vtt at all), maybe it could be documented, but anyway that kind of features is so special that people don't reasonably expect the documentation to be fully exhaustive (would certainly be overly long and never read).

PR #423, opened 2023-12-17
Add support for document filters

### LINKED ISSUE #417: STL reader: option to set a generic, basic style
The EBU mapping is not always welcome. For simple distribution to end users, a more generic/simplified format appears to be sometimes a better choice.
The first patch disables the "ebu style".
The second disables the vertical positioning in favor of a fixed bottom position, with an adjustable margin.

See https://github.com/sandflow/ttconv/pull/416#issue-2002605938

palemieux:
`tt convert -i <.scc file> -o <.ttml file> --itype SCC --otype TTML --filter lcd --config '{"general": {"progress_bar":false, "log_level":"WARN"}, "lcd": {"bg_color": "blue"}}'`

Closes #417


ngaullier:
README: double quote closing missing after "black on
Examples: "#FF0000" (red), "transparent", "black
I think you can just add a README entry for lcd/ "preserve_text_align" and everything looks very good to me! Thank you!
There is still a last late thing, I apologize I just get it now, it is the safe area : some users might want a different safe area on the height and on the width. My production shows 10% in width and 0% in height: I guess that on a smartphone, the screen height is so small (and with no overscan) that it is nice to have subtitles stick to the bottom, but on the other side, having the eyes moving hard left to hard right all the time must be annoying. I don't know if this opinion is shared or not, I just can tell you this is common practice for a (very) big content owner here; I don't know how people would react if having to change to an "isotropic" safe area.
I mean, you have done such a work, it does not seem reasonable to me to ask you even more! So up to you!

palemieux:
Thanks for the eagle eye on the readme. Should be fixed now.

Re: safe area, I think I need to better understand the use case to have an opinion -- happy to have a side discussion on the topic. At a high-level, I do not think safe area should be set to accommodate a specific set of devices. Device limitation/idiosyncrasies should be handled at the device by the player, with the understanding that some reasonable safe area exists. Makes sense?

In the meantime, I suggest we merge the PR as-is and iterate, if needed, before the next release.

ngaullier:
OK, that's fine.

PR #427, opened 2024-05-13
SCC: Handle non-matched SCC codes in SccWord.get_channel

### LINKED ISSUE #428: SCC: is_code() can return `true` even if the word is not a code
From @atsampson:

https://github.com/sandflow/ttconv/blob/536f1c8d16793299cffee9e95aa3b325ed3281c3/src/main/python/ttconv/scc/word.py#L115

is_code() can return True even if self.code is None, i.e., no code was found in the constructor.

See https://github.com/sandflow/ttconv/pull/427

atsampson:
If is_code returns True but _find_code doesn't match the code, get_channel would crash rather than returning None.

(This was another case of a somewhat-mangled CC stream from a noisy LaserDisc via ld-decode; the invalid words were 1f38, 1d00, 1432, 1000.)

Closes #428

### DIFF
--- a/src/main/python/ttconv/scc/word.py
+++ b/src/main/python/ttconv/scc/word.py
@@ -119,8 +119,8 @@ class SccWord:
 
   def get_channel(self) -> Optional[SccChannel]:
     """Returns the caption channel, if the word is an SCC code"""
-    if self.is_code():
-      if isinstance(self.code, SccPreambleAddressCode):
-        return self.code.get_channel()
-      return self.code.get_channel(self.value)
-    return None
+    if self.code is None:
+      return None
+    if isinstance(self.code, SccPreambleAddressCode):
+      return self.code.get_channel()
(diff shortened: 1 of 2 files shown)


palemieux:
Feels like `is_code()` should return `False` if `self.code` is `None`.

atsampson:
That was my first thought, but it's used in `_find_code` which is called from the constructor. Maybe `_find_code` should just try and match regardless?

palemieux:
I see now, and agree that your patch fixes the immediate issue.

I see a broader issue with the semantics of `is_code()` and `get_code()` but probably should be fixed as a separate issue.

Can you add a test case that covers the issue that triggered the PR, and then we can merge.

Thanks!

palemieux (approved):

atsampson:
I've added a unit test for `SccWord.get_channel`, and another that tests that all the methods return something for all possible codes.

PR #476, opened 2025-12-30
475: VTT reader does not display cues with negative line attribute values

### LINKED ISSUE #475: VTT reader does not display cues with negative "line" attribute values
Some VTT files have a negative "line" value, e.g. the following cue:

00:00:01.320 --> 00:00:03.520 align:center line:-1
Cue 1 on the last line

This line should be displayed at the bottom of the area, but with the current code it does not show, because the computed line position and line extent are bad.

It seems negative "line" values have been thought of:
https://github.com/sandflow/ttconv/blob/3788df9bf7555b7b4789b6e737174a839c32794b/src/main/python/ttconv/vtt/reader.py#L295C75-L295C109
But this code actually results in a line position that is greater than 100%, and also a line extent that is negative.

I think a simple way to fix this is to change the '-' sign to '+' as "100 + 100 * line_num/_DEFAULT_ROWS" gives the good offset.

pcochenn-harmonicinc:
Integer line values can be positive or negative. When positive, lines are counted from the top of the screen, and when negative, from the bottom of the screen.

The code did not handle negative values correctly, as line positions ended up greater than 100(%), hence out of the screen.
The fix handles negative values properly and adds a test case for both positive/negative integers and percentage.

Note that line value 0 should probably be handled as a positive value, but I did not change the current logic to remain backward compatible.

Closes #475

### DIFF
--- a/src/main/python/ttconv/vtt/reader.py
+++ b/src/main/python/ttconv/vtt/reader.py
@@ -292,9 +292,9 @@ def _get_or_make_region(
       line_num = parse_vtt_int(value[0])
       if line_num is not None:
         if writing_mode in (styles.WritingModeType.rltb, styles.WritingModeType.lrtb):
-          line_offset = 100 * line_num/_DEFAULT_ROWS if line_num > 0 else 100 + 100 * line_num/_DEFAULT_ROWS
+          line_offset = 100 * line_num/_DEFAULT_ROWS if line_num >= 0 else 100 + 100 * line_num/_DEFAULT_ROWS
         else:
-          line_offset = 100 * line_num/_DEFAULT_COLS if line_num > 0 else 100 + 100 * line_num/_DEFAULT_COLS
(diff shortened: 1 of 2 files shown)


palemieux:
> Note that line value 0 should probably be handled as a positive value, but I did not change the current logic to remain backward compatible.

It looks like this should also be fixed since the [spec](https://www.w3.org/TR/webvtt1/#cue-settings) states the following. I would fix it here unless you think a separate issue is necessary.

>  Positive line numbers count from the start of the video viewport (the first line is numbered 0), negative line numbers from the end of the video viewport (the last line is numbered −1).

palemieux (changes requested):
As discussed at https://github.com/sandflow/ttconv/pull/476#issuecomment-3699957771 , I suggest fixing the `line: 0` bug here as well unless you think a separate issue should be opened.

PR #482, opened 2026-02-16
Add IMSC 1.1 validating filter

### LINKED ISSUE #484: Add IMSC 1.1 content profile signaling to WebVTT reader
The ttconv data model extends the IMSC 1.1 data model slightly, e.g., allows negative lengths on `origin`, closely. As a result it is not possible to generally guarantee that the TTML writer will generate a document conformant to IMSC 1.1. This issue will grow more acute as the model evolves to support future versions of IMSC that add features, e.g., IMSC 1.3 with superscript and subscript.

While the conformance to IMSC 1.1 of the output of ttconv should always be determined by an actual validator, it should be possible to avoid significant incompatibility by implementing a filter that fails when encountering parameter values and elements that are not supported by IMSC 1.1.

See also #369

eko:
Hi @palemieux,

Following this pull request: https://github.com/sandflow/ttconv/pull/461

We need to have the content profiles signaling also when the input file is a VTT file.

Could you please review this pull request?

Thank you

Closes #484

### DIFF
--- a/src/main/python/ttconv/scc/caption_paragraph.py
+++ b/src/main/python/ttconv/scc/caption_paragraph.py
@@ -497,17 +497,12 @@ class _SccParagraphRegion:
     # Convert origin cells to percentages
     if self._paragraph.get_caption_style() is SccCaptionStyle.RollUp:
       # The region origin x offset
-      region_origin = get_position_from_offsets(paragraph_origin.x.value, max(0, self._top))
+      region_origin = get_position_from_offsets(paragraph_origin.x.value, self._top)
       region_origin_pct = convert_cells_to_percentages(region_origin, self._doc.get_cell_resolution())
     else:
       # The region origin matches with paragraph origin
       region_origin_pct = convert_cells_to_percentages(paragraph_origin, self._doc.get_cell_resolution())
 
-    # Clamp to non-negative values (section 8.4.5)
-    region_origin_pct = CoordinateType(
-      x=LengthType(value=max(0, region_origin_pct.x.value), units=region_origin_pct.x.units),
-      y=LengthType(value=max(0, region_origin_pct.y.value), units=region_origin_pct.y.units),
-    )
     region.set_style(StyleProperties.Origin, region_origin_pct)
 
     # The region width is initialized with he paragraph width (up to the right of the safe area)
--- a/src/main/python/ttconv/style_properties.py
+++ b/src/main/python/ttconv/style_properties.py
@@ -541,9 +541,7 @@ class StyleProperties:
     def validate(value: ExtentType):
       return isinstance(value, ExtentType) \
         and value.width.units in (LengthType.Units.pct, LengthType.Units.px, LengthType.Units.rw)  \
-        and value.height.units in (LengthType.Units.pct, LengthType.Units.px, LengthType.Units.rh) \
-        and value.width.value >= 0 \
-        and value.height.value >= 0
+        and value.height.units in (LengthType.Units.pct, LengthType.Units.px, LengthType.Units.rh)
 
 
   class FillLineGap(StyleProperty):
@@ -588,7 +586,7 @@ class StyleProperties:
 
     @staticmethod
     def validate(value):
-      return isinstance(value, LengthType) and value.value >= 0
+      return isinstance(value, LengthType)
 
 
   class FontStyle(StyleProperty):
@@ -633,8 +631,7 @@ class StyleProperties:
 
     @staticmethod
     def validate(value):
-      return value == SpecialValues.normal or \
-        (isinstance(value, LengthType) and value.units != LengthType.Units.c and value.value >= 0)
+      return value == SpecialValues.normal or (isinstance(value, LengthType) and value.units != LengthType.Units.c)
 
 
   class LinePadding(StyleProperty):
@@ -650,8 +647,7 @@ class StyleProperties:
     @staticmethod
     def validate(value: LengthType):
       return isinstance(value, LengthType) and \
-        value.units in (LengthType.Units.c, LengthType.Units.rh, LengthType.Units.rw) and \
-        value.value >= 0
+        value.units in (LengthType.Units.c, LengthType.Units.rh, LengthType.Units.rw)
 
 
   class LuminanceGain(StyleProperty):
@@ -716,9 +712,7 @@ class StyleProperties:
     def validate(value: CoordinateType):
       return isinstance(value, CoordinateType) \
         and value.x.units in (LengthType.Units.pct, LengthType.Units.px, LengthType.Units.rw)  \
-        and value.y.units in (LengthType.Units.pct, LengthType.Units.px, LengthType.Units.rh) \
-        and value.x.value >= 0 \
-        and value.y.value >= 0
+        and value.y.units in (LengthType.Units.pct, LengthType.Units.px, LengthType.Units.rh)
 
 
   class Overflow(StyleProperty):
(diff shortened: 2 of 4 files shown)


palemieux:
@eko Thanks for making progress on this issue. Instead of a _validator_ (which is a much more complex beast), I suggest refactoring this PR into an _imsc11_ filter (see the `lcd` filter for an example). The filter would have the following characteristics:

* fail immediately if any of the IMSC 1.1 constraints is broken (no need to log each of the deviations, which are meaningless to the end-user and the results of the internal machinery)
* add the IMSC 1.1 Text Profile designator to the `ttp:contentProfiles` if there are no failures

This would allow the user to insert the filter to any conversion.

Ok? If so, can you take a stab at the refactoring?

The filter should be unit-tested against the IMSC test suite at `src/test/resources/ttml` (I can handle that part if you prefer).

src/main/python/ttconv/imsc/imsc_11_text_profile_validator.py
```diff
@@ -0,0 +1,266 @@
...
+  model.Ruby,
+  model.Rb,
+  model.Rt,
+  model.Rp,
+  model.Rbc,
+  model.Rtc,
+  model.Region,
+)
+
+_IMSC_11_TEXT_STYLE_PROPS = frozenset(imsc_styles.StyleProperties.BY_MODEL_PROP.keys())
```
palemieux: This should be an explicit list since IMSC 1.3 will add support for superscript/subscript.

src/main/python/ttconv/imsc/imsc_11_text_profile_validator.py
```diff
@@ -0,0 +1,266 @@
...
+def _validate_region(
+  violations: typing.List[str],
+  region: model.Region,
+):
+  '''Validates region-specific constraints.'''
+
+  _validate_element(violations, region)
+
+  # Section 8.4.2: tts:extent SHALL be present on all region elements
+  if region.get_style(styles.StyleProperties.Extent) is None:
```
palemieux: This only retrieves the specified style. [Initial styling](https://www.w3.org/TR/ttml2/#styling-vocabulary-initial) must also be retrieved using `get_initial_value()` on the document.

In fact, style values can also be inherited (for those that are inheritable) or set through the `set` element.

A more robust approach would be to first generate the ISD sequence and perform the checks on each ISD (where the styles have been completely resolved).

eko: Good point. The validation now uses a hybrid approach.

This way, unit-based constraints are checked where the unit info is still available, and value-based constraints are checked on the ISD where styles are fully resolved.

src/main/python/ttconv/filters/doc/imsc11text.py
```diff
@@ -0,0 +1,413 @@
...
+])
+
+# Section 8.4.5: properties that allow negative lengths.
+_NEGATIVE_LENGTH_ALLOWED_PROPS = frozenset([
+  styles.StyleProperties.Disparity,
+  styles.StyleProperties.TextShadow,
+])
+
+# Section 7.12.9: rh and rw units SHALL only appear on these properties.
+_RH_RW_ALLOWED_PROPS = frozenset([
```
palemieux: `rh` and `rw` are allowed on all properties. 7.2.19 applies only to `region` elements.

src/main/python/ttconv/filters/doc/imsc11text.py
```diff
@@ -0,0 +1,413 @@
...
+  and Origin."""
+  if style_prop not in _RH_RW_ALLOWED_PROPS and \
+    length.units in (styles.LengthType.Units.rh, styles.LengthType.Units.rw):
+    raise ValueError(
+      f"{context} uses '{length.units.value}' unit, "
+      f"rh/rw only allowed on Extent, Position, Origin (section 7.12.9)"
+    )
+
+
+def _check_length_constraints(element, style_prop, value):
```
palemieux: `c` length units are allowed only on `linePadding` (7.2.18).

src/main/python/ttconv/filters/doc/imsc11text.py
```diff
@@ -0,0 +1,413 @@
...
+    _check_isd_negative(elem_name, line_height, "LineHeight")
+
+  padding = element.get_style(styles.StyleProperties.Padding)
+  if padding is not None:
+    _check_isd_negative(elem_name, padding.before, "Padding.before")
+    _check_isd_negative(elem_name, padding.end, "Padding.end")
+    _check_isd_negative(elem_name, padding.after, "Padding.after")
+    _check_isd_negative(elem_name, padding.start, "Padding.start")
+
+  text_outline = element.get_style(styles.StyleProperties.TextOutline)
```
palemieux: Enforce https://www.w3.org/TR/ttml-imsc1.1/#text-textOutline-unblurred

src/main/python/ttconv/filters/doc/imsc11text.py
```diff
@@ -0,0 +1,413 @@
...
+    # ContentDocument specified-style checks on body elements
+    body = doc.get_body()
+    if body is not None:
+      for element in body.dfs_iterator():
+        _validate_element_styles(element)
+
+    # ISD-based resolved style checks
+    isd_sequence = ISD.generate_isd_sequence(doc)
+    for _, isd in isd_sequence:
+      for isd_region in isd.iter_regions():
```
palemieux: Validate https://www.w3.org/TR/ttml-imsc1.1/#maximum-number

src/main/python/ttconv/filters/doc/imsc11text.py
eko: I added check in `_validate_isd_element`: on resolved ISD elements, if `textOutline.thickness` exceeds 10% of fontSize (when both are in the same units), the filter raises.

src/main/python/ttconv/filters/doc/imsc11text.py
```diff
@@ -0,0 +1,413 @@
...
+def _check_isd_negative(elem_name, value, context):
+  """Checks a single resolved length for negative value."""
+  if isinstance(value, styles.LengthType) and value.value < 0:
+    raise ValueError(
+      f"{elem_name}: resolved {context} is negative "
+      f"({value.value}), not allowed per section 8.4.5"
+    )
+
+
+def _validate_isd_element(element):
```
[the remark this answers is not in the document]
eko: I added defense-in-depth check in `_validate_isd_element`: rubyAlign must be center or spaceAround. The model's RubyAlignType enum already constrains this, but the explicit check provides additional safety.

src/main/python/ttconv/filters/doc/imsc11text.py
```diff
@@ -0,0 +1,413 @@
...
+    body = doc.get_body()
+    if body is not None:
+      for element in body.dfs_iterator():
+        _validate_element_styles(element)
+
+    # ISD-based resolved style checks
+    isd_sequence = ISD.generate_isd_sequence(doc)
+    for _, isd in isd_sequence:
+      for isd_region in isd.iter_regions():
+        for element in isd_region.dfs_iterator():
```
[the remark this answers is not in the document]
eko: I added `_check_px_requires_extent_root()` that scans all specified styles for px units, and if found, checks `doc.has_initial_value(StyleProperties.Extent)`.

Limitation: the model stores `tts:extent` on tt as px_resolution rather than as an initial value, so this check won't catch documents where tts:extent was set on tt through the IMSC reader. This is the best we can do at the filter level.

eko:
Thank you for your kindly support @palemieux!

Could you please create a new tag so we will ensure this version will remain stable?

palemieux:
> Could you please create a new tag so we will ensure this version will remain stable?

We could release a beta version if it is easier than referencing a commit.

This is a good time to try a large body of documents and make they work.

I expect a release candidate in two weeks.

In the meantime, I  plan to improve the WebVTT reader and enforce positive length requirements in the model: this PR has highlighted issues with WebVTT cue positioning.

Expect further changes.

PR #486, opened 2026-03-01
* Impose positive length constraints on data model to match IMSC

### LINKED ISSUE #483: Validate length polarity at StyleProperties
Length-based style properties at `StyleProperties` should validate polarity requirements per the [IMSC data model](https://www.w3.org/TR/ttml-imsc1.1).

### LINKED ISSUE #487: WebVTT reader generates regions that extends past the root container
The cue settings `position:20%,center` results in a region with an origin at `x = -27.5%`.

palemieux:
* Fix WebVTT reader positioning algorithm

Closes #483
Closes #487

### DIFF
--- a/src/main/python/ttconv/style_properties.py
+++ b/src/main/python/ttconv/style_properties.py
@@ -69,7 +69,7 @@ class LengthType:
     if not isinstance(self.units, LengthType.Units):
       raise ValueError("Invalid units")
 
-  def is_positive(self):
+  def is_non_negative(self):
     '''True if the length value is positive
     '''
     return self.value >= 0
@@ -547,7 +547,7 @@ class StyleProperties:
       return isinstance(value, ExtentType) \
         and value.width.units in (LengthType.Units.pct, LengthType.Units.px, LengthType.Units.rw)  \
         and value.height.units in (LengthType.Units.pct, LengthType.Units.px, LengthType.Units.rh) \
-        and value.width.is_positive() and value.height.is_positive()
+        and value.width.is_non_negative() and value.height.is_non_negative()
 
 
   class FillLineGap(StyleProperty):
@@ -592,7 +592,7 @@ class StyleProperties:
 
     @staticmethod
     def validate(value):
-      return isinstance(value, LengthType) and value.is_positive()
+      return isinstance(value, LengthType) and value.is_non_negative()
 
 
   class FontStyle(StyleProperty):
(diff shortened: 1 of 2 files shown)


src/main/python/ttconv/style_properties.py
```diff
@@ -69,6 +69,11 @@ def __post_init__(self):
     if not isinstance(self.units, LengthType.Units):
       raise ValueError("Invalid units")
 
+  def is_positive(self):
```
eko: 💡 `is_positive()` returns `self.value >= 0`, which is mathematically "non-negative", not "positive". `is_non_negative()` would be clearer.

src/main/python/ttconv/style_properties.py
```diff
@@ -662,8 +669,7 @@ def make_initial_value():
 
     @staticmethod
     def validate(value: Numeric):
-      return isinstance(value, numbers.Number)
-
+      return isinstance(value, numbers.Number) and value > 0.0
```
eko: 💡 I am not sure about this one but TTML spec says that luminanceGain is a non-negative number and here it's strictly positive, maybe this value should be `>= 0.0`?

https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-attribute-luminanceGain

palemieux: `tts:luminanceGain == 0` is absurd, but, as you say, syntactically permitted in TTML 2!

src/main/python/ttconv/vtt/reader.py
```diff
@@ -223,188 +227,273 @@ def parse_vtt_int(value: str):
...
+    r.writing_mode = styles.WritingModeType.tbrl
   elif value is not None:
     LOGGER.warning("Bad vertical setting value: %s", value)
 
+  # cue text align
+  cue_text_align = "center"
+  value = cue_settings.get("align")
+  if value in ("left", "right", "start", "center", "end"):
+    cue_text_align = value
+  else:
```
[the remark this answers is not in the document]
palemieux: Addressed 7595386bd2c4082eb7434d8c8a5c5132a1124281

src/main/python/ttconv/style_properties.py
palemieux: Addressed at 7595386bd2c4082eb7434d8c8a5c5132a1124281

palemieux:
@eko Thanks so much for the review. I plan to release a beta version shortly.

Issue #354, opened 2021-12-02
ttconv should consider EBU STL format as an output


Vincent-Dabouineau:
A writer really nice to have. It is difficult to handle / modify a binary format as STL directly. It is easier to manipulate the canonical or TTML files within an automatic subtitle resync process for e.g, then output to the EBU STL format, still compliant with the major broadcast workflows at this time.

xchange11:
> A writer really nice to have. It is difficult to handle / modify a binary format as STL directly. It is easier to deal with the canonical or TTML files (within an automatic subtitling resync process for e.g), then output to the EBU STL « mezzanine » format for next purposes, compliant with the main broadcast workflows at this time.

@Vincent-Dabouineau Yes, this kind of feature is really helpful and I also thought of this. But as @palemieux mentioned in another issue every input is converted into the internal generic model and then again to all possible output formats. This has to implications:

1. An EBU STL writer needs also the handle other input formats like IMSC that have a different or broader feature scope. 
2. Currently all metadata is discarded.

For 1. you may do something similar was what is currently applied to the SRT writer: you only use very basic features like time, text and colours. But regarding 2.: do you also want to preserve the metadata in the GSI section of the EBU-STL file?

xchange11:
For the time being, you may want to have a look at the [Subtitle Conversion Framework](https://github.com/IRT-Open-Source/scf/). You can translate EBU STL with the [STL2STLXML](https://github.com/IRT-Open-Source/scf/tree/master/modules/STL2STLXML) module to an XML representation of STL, manipulate the STLXML and then convert it back to EBU-STL with the [STLXML2STL](https://github.com/IRT-Open-Source/scf/tree/master/modules/STLXML2STL) module. 

If the processing of the STLXML is too cumbersome you can also translate it to EBU-TT with the [STLXML2EBU-TT](https://github.com/IRT-Open-Source/scf/tree/master/modules/STLXML2EBU-TT) module, process this TTML form, convert it back to STLXML with the [EBU-TT2STLXML](https://github.com/IRT-Open-Source/scf/tree/master/modules/EBU-TT2STLXML) module and use as above the [STLXML2STL](https://github.com/IRT-Open-Source/scf/tree/master/modules/STLXML2STL) module as last step.

This workflow is used in some operational broadcast workflows.

PR #505, opened 2026-04-21
Fix STL30.01 DFC mapping to use 30fps & add STL29.01

### LINKED ISSUE #504: STL 30.01 disc format code should map to 30 FPS not 29.97 FPS
The STL GSI block parser in stl/datafile.py maps the `STL30.01` Disk Format Code to `Fraction(30000, 1001)` - i.e. 29.97 fps

https://github.com/sandflow/ttconv/blob/e024e34f78ffee2eed70dc5defbaa4e6915990cb/src/main/python/ttconv/stl/datafile.py#L162-L168 

According to the EBU Tech 3264 specification for STL (section 4.2.2), `STL30.01` is a standard Disc Format Code in STL and means 30 frames/s: https://tech.ebu.ch/docs/tech/tech3264.pdf

Therefore correct mapping should be `Fraction(30, 1)`, while the non-standard extension to support 29.97 FPS files should use `STL29.01` (e.g. as used in [Subtitle Edit](https://github.com/SubtitleEdit/subtitleedit/blob/main/src/UI/Features/Files/ExportEbuStl/ExportEbuStlViewModel.cs#L93))

powelli:
Per EBU Tech 3264 section 4.2.2, `STL30.01` specifies 30 FPS, not 29.97 FPS.

This change corrects the mapping and adds STL29.01 → 29.97 FPS as a non-standard extension for files that use that convention (e.g. as used by Subtitle Edit).

Fixes #504

### DIFF
--- a/src/main/python/ttconv/stl/datafile.py
+++ b/src/main/python/ttconv/stl/datafile.py
@@ -163,7 +163,8 @@ _DFC_FRACTION_MAP = {
   b'STL23.01': Fraction(24000, 1001),
   b'STL24.01': Fraction(24),
   b'STL25.01': Fraction(25),
-  b'STL30.01': Fraction(30000, 1001),
+  b'STL29.01': Fraction(30000, 1001),
+  b'STL30.01': Fraction(30),
   b'STL50.01': Fraction(50)
 }
 


palemieux:
@eko Any strong opinion?

eko (commented):
Hi, it appears that using 30 frames per second is appropriate (https://tech.ebu.ch/docs/tech/tech3264.pdf, paragraph 4.2.2).

However, I’m not sure how this affects on STL 29.01.

powelli:
@eko Thanks for the review. I realise that any codes other than `STL25.01` & `STL30.01` are non-standard, but I figure that adding `29.01` is appropriate as:
1. The library already supports other non-standard codes like `STL23.01`
2. I know from practice that some systems do use non-standard framerates to match their video content
3. Other packages like Subtitle Edit use this non-standard value to represent 29.97 FPS
4. It may be useful if an STL writer were ever implemented in the future, as suggested in #354

palemieux:
See https://github.com/sandflow/ttconv/issues/504#issuecomment-4299341700.

Have you ever come across actual STL content labeled `STL30.01` that is conformed to actual 30 fps video?

I am tempted to say that both `STL30.01` and `STL29.01` mean 29.97 fps video.

powelli:
Closing as official standards indicate that the association with 29.97 FPS is correct, see comment: https://github.com/sandflow/ttconv/issues/504#issuecomment-4305323006

PR #506, opened 2026-04-22
Fix SCC NDF timecode drift: parse at FPS_29_97, fix add_frames() DF flag

### LINKED ISSUE #507: SCC reader incorrectly assumes that timecode is at 30 FPS
https://github.com/sandflow/ttconv/blob/a426e4e0faef61b07820e4c4098c2129c203491f/src/main/python/ttconv/scc/line.py#L80

SCC files can only be 29.97 fps.

See also https://github.com/sandflow/ttconv/pull/506

### LINKED ISSUE #508: Timecode add_frames() does not handle NDF timecode
https://github.com/sandflow/ttconv/blob/a426e4e0faef61b07820e4c4098c2129c203491f/src/main/python/ttconv/time_code.py#L214

should be:

`new_time_code = SmpteTimeCode.from_frames(frames, self._frame_rate, self._is_drop_frame)`

See also https://github.com/sandflow/ttconv/pull/506

powelli:
This change addresses 2 related issues:

1. SCC files were parsed assuming 30 FPS instead of 29.97 FPS, causing ~3.6s/hour of timing drift for Non-Drop Frame (NDF) content. Drop Frame (DF) content was unaffected because `SmpteTimeCode.parse()` already applied a 1000/1001 rate correction for DF timecodes, which happened to compensate for the wrong input rate.

2. `add_frames()` re-derived the DF/NDF flag from the frame rate denominator instead of preserving it from the original timecode, silently converting NDF 29.97 timecodes to drop-frame after the first frame advance.

## References
- [Telestream](https://support.telestream.net/s/article/SCC-Files-and-23-98-fps-or-Other-Frame-Rates)
- [theneitherworld.com](http://www.theneitherworld.com/mcpoodle/SCC_TOOLS/DOCS/SCC_FORMAT.HTML)

## Individual changes
- **`scc/line.py`** - Changed `FPS_30` to `FPS_29_97` when parsing SCC line timecodes. SCC (CEA-608) files carry no embedded framerate & 29.97 fps is the correct default (NTSC framerate as defined in CEA-608).

- **`time_code.py`** - `add_frames()` now passes `self._is_drop_frame` explicitly to `from_frames()`. Previously it defaulted `is_df` from the rate denominator, silently converting NDF FPS_29_97 timecodes to drop-frame on every frame advance as this was incorrectly based on the framerate denominator.

- **`test_scc_line.py`** - Existing timecode assertions updated from `FPS_30` to `FPS_29_97` (with explicit `is_df=False` for NDF). New paired tests added asserting NDF and DF SCC line timecodes parse to the correct frame rate and drop-frame flag.

4. **`test_scc_reader.py`** - `check_element_timecode()` helper updated to parse expected timecodes at `FPS_29_97`. Per-frame span offset assertions updated from `1/30 * N` to `1/29.97 * N`.

Closes #508 
Closes #507

### DIFF
--- a/src/main/python/ttconv/scc/line.py
+++ b/src/main/python/ttconv/scc/line.py
@@ -42,7 +42,7 @@ from ttconv.scc.codes.special_characters import SccSpecialCharacter
 from ttconv.scc.context import SccContext
 from ttconv.scc.disassembly import get_scc_word_disassembly
 from ttconv.scc.word import SccWord
-from ttconv.time_code import SmpteTimeCode, FPS_30
+from ttconv.time_code import SmpteTimeCode, FPS_29_97
 
 LOGGER = logging.getLogger(__name__)
 
@@ -77,7 +77,7 @@ class SccLine:
       return None
 
     time_code = match.group(1)
-    time_offset = SmpteTimeCode.parse(time_code, FPS_30)
+    time_offset = SmpteTimeCode.parse(time_code, FPS_29_97)
 
     hex_words = line.split('\t')[1].split(' ')
     scc_words = [SccWord.from_str(hex_word) for hex_word in hex_words if hex_word]
--- a/src/main/python/ttconv/time_code.py
+++ b/src/main/python/ttconv/time_code.py
@@ -211,7 +211,7 @@ class SmpteTimeCode(_HHMMSSTimeExpression):
     """Add frames to the current time code"""
     frames = self.to_frames() + nb_frames
 
-    new_time_code = SmpteTimeCode.from_frames(frames, self._frame_rate)
+    new_time_code = SmpteTimeCode.from_frames(frames, self._frame_rate, self._is_drop_frame)
 
     self._hours = new_time_code.get_hours()
     self._minutes = new_time_code.get_minutes()
--- a/src/test/python/test_scc_line.py
+++ b/src/test/python/test_scc_line.py
@@ -30,16 +30,36 @@ import unittest
 
 from ttconv.scc.line import SccLine
 from ttconv.scc.caption_style import SccCaptionStyle
-from ttconv.time_code import SmpteTimeCode, FPS_30
+from ttconv.time_code import SmpteTimeCode, FPS_29_97
 
 
 class SccLineTest(unittest.TestCase):
+
+  def test_scc_line_from_str_ndf_timecode_uses_fps_29_97(self):
+    """NDF SCC timecodes (colon-delimited) are parsed at 29.97 fps, non-drop-frame."""
+    line_str = "01:00:00:00	9420 9420"
+    scc_line = SccLine.from_str(line_str)
+    self.assertIsNotNone(scc_line)
+    self.assertEqual(FPS_29_97, scc_line.time_code.get_frame_rate())
+    self.assertFalse(scc_line.time_code.is_drop_frame())
+    # 01:00:00:00 NDF at 29.97 fps = ~3603.6s (slightly longer than clock hour due to 1/29.97 frame duration)
+    self.assertEqual(SmpteTimeCode(1, 0, 0, 0, FPS_29_97, False).to_temporal_offset(), scc_line.time_code.to_temporal_offset())
+
+  def test_scc_line_from_str_df_timecode_is_drop_frame(self):
+    """DF SCC timecodes (semicolon-delimited) are parsed as drop-frame 29.97 fps."""
+    line_str = "01:00:00;00	9420 9420"
+    scc_line = SccLine.from_str(line_str)
+    self.assertIsNotNone(scc_line)
+    self.assertEqual(FPS_29_97, scc_line.time_code.get_frame_rate())
+    self.assertTrue(scc_line.time_code.is_drop_frame())
+    # 01:00:00;00 DF at 29.97 fps = ~3600.0s (drop-frame compensates back to real clock time)
+    self.assertEqual(SmpteTimeCode(1, 0, 0, 0, FPS_29_97, True).to_temporal_offset(), scc_line.time_code.to_temporal_offset())
 
   def test_scc_line_from_str_with_pop_on_style(self):
     line_str = "01:03:27:29	94ae 94ae 9420 9420 94f2 94f2 c845 d92c 2054 c845 5245 ae80 942c 942c 8080 8080 942f 942f"
     scc_line = SccLine.from_str(line_str)
     self.assertEqual(18, len(scc_line.scc_words))
-    self.assertEqual(SmpteTimeCode(1, 3, 27, 29, FPS_30).to_temporal_offset(), scc_line.time_code.to_temporal_offset())
+    self.assertEqual(SmpteTimeCode(1, 3, 27, 29, FPS_29_97, False).to_temporal_offset(), scc_line.time_code.to_temporal_offset())
     self.assertEqual("01:03:27:29	{ENM}{ENM}{RCL}{RCL}{1504}{1504}HEY, THERE.{EDM}{EDM}{}{}{EOC}{EOC}", scc_line.to_disassembly())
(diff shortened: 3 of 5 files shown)


powelli:
Note - this change will correct drift for SCC _standard_ 29.97 FPS NDF timecodes (and as a by-product improve the situation for other non-integer framerates with NDF timecodes such as 23.976 NDF).

For SCC files mastered against _non-standard_ integer framerates (such as 25 FPS or 30 FPS) this change will actually introduce some discrepancy.

A follow-up PR is planned to add an optional framerate parameter to the SCC reader configuration (similar to the SCC writer configuration) to allow this behaviour to be overridden when using non-standard framerates.

palemieux:
Did you consider the `frame_rate` config option?

Also, do you have a sample file?

See also https://github.com/sandflow/ttconv/pull/458

powelli:
> Did you consider the `frame_rate` config option?
> 
> Also, do you have a sample file?
> 
> See also #458

Thanks for the review @palemieux 

I have a second branch (based off this on) in which I plan to add the `frame_rate` option for the SCC Reader, as it's currently only supported for the SCC Writer. I raised this first as I wanted to align the correct default frame rate value first so that I wasn't trying to change both behaviours in a single PR.

I can provide a sample file, although it won't really differ from the existing SCC test assets in the repo in any meaningful way. Is there anything specific you'd like me to try and show?

palemieux:
> SCC files were parsed assuming 30 FPS instead of 29.97 FPS, causing ~3.6s/hour of timing drift for Non-Drop Frame (NDF) content. 

Can you provide an example of a SCC timestamp and the corresponding expected temporal offset in seconds you expect?

powelli:
> > SCC files were parsed assuming 30 FPS instead of 29.97 FPS, causing ~3.6s/hour of timing drift for Non-Drop Frame (NDF) content.
> 
> Can you provide an example of a SCC timestamp and the corresponding expected temporal offset in seconds you expect?

Sure! Here's a simple fixture I've used for validation:
```
Scenarist_SCC V1.0

00:00:00:00	9420 9420 94d0 94d0 5380 942f 942f

01:00:00:29	9420 9420 94d0 94d0 4580 942f 942f

01:01:00:00	942c 942c

```

A few notes on this fixture:
1. It's using NDF timecodes (`HH:MM:SS:FF` rather than `HH:MM:SS;SS`)
2. It's based on 29.97 NDF content, the highest expected FF value for which is 29
3. The line at `01:00:00:29` is the one of interest, as this demonstrates the accumulated drift from the incorrect FPS

Disassembled fixture:
```
00:00:00:00	{CC1|RCL}{CC1|RCL}{CC1|1400}{CC1|1400}S{CC1|EOC}{CC1|EOC}
01:00:00:29	{CC1|RCL}{CC1|RCL}{CC1|1400}{CC1|1400}E{CC1|EOC}{CC1|EOC}
01:01:00:00	{CC1|EDM}{CC1|EDM}
```

That line has 5 words prior to the `EOC` display code, which is the point at which the caption will be displayed and the timestamp that will be recorded in the temporal offset.

So the _**expected**_ temporal offset at 29.97 NDF - at which each timecode second is 1001/1000 (1.001) wall-clock seconds works out at:

```
elapsed_tc_seconds = (HH * 3600 + MM * 60 + SS) * (1001/1000)
elapsed_tc_seconds = (1 * 3600 + 0 * 60 + 0) * (1001/1000)
elapsed_tc_seconds = 3600 * (1001/1000)
elapsed_tc_seconds = 3603.6

frame_seconds = (FF + 5) * 1/FPS
frame_seconds = (29 + 5) * 1/(30*1000/1001)
frame_seconds = (29 + 5) * (1001/30000)
frame_seconds ~= 1.1344666667

temporal_offset = elapsed_tc_seconds + frame_seconds
temporal_offset ~= 3,604.7344666667 seconds
```

If we instead assume 30 FPS (as the library does prior to this change) - each timecode second is considered exactly 1 wall-clock second, and it works out at:
```
elapsed_tc_seconds = (HH * 3600 + MM * 60 + SS) * 1
elapsed_tc_seconds = (1 * 3600 + 0 * 60 + 0) * 1
elapsed_tc_seconds = 3600 * 1
elapsed_tc_seconds = 36000

frame_seconds = (FF + 5) * 1/FPS
frame_seconds = (29 + 5) * 1/(30)
frame_seconds = (29 + 5) * 1/30
frame_seconds ~= 1.1333333333

temporal_offset = elapsed_tc_seconds + frame_seconds
temporal offset ~= 3,601.1333333333
```

This means that the caption will be displayed roughly _**3.6 seconds too early**_ - which is significant for sync

palemieux:
@powelli Ok. I created an issue at https://github.com/sandflow/ttconv/issues/507. I agree that SCC files are always 29.97 fps.

Can we tackle the `add_frames()` issue separately since it does not apply for the SCC reader and requires additional unit tests. 

P.S.: Not sure how this bug could have been around for so long! Maybe most SCC files have DF timecode.

powelli:
Thanks @palemieux - I'm not sure if we can separate the 2 though. As we have both DF & NDF usage of a non-integer framerate the fix for `add_frames()` becomes critical.

We first infer DF/NDF when parsing the timecode at https://github.com/sandflow/ttconv/blob/master/src/main/python/ttconv/scc/line.py#L80 - based on the timecode patterns you introduced in #458 

`add_frames()` is used as the reader steps through each word in an SCC line. If it _doesn't_ pass the `self._is_drop_frame` flag through to `from_frames()`, then this latter function will incorrectly default to inferring it from the framerate denominator at https://github.com/sandflow/ttconv/blob/master/src/main/python/ttconv/time_code.py#L260

Given we're now using a non-integer framerate `FPS_29_97 = Fraction(30000, 1001)` but **non**-drop frame, this inference will be incorrect and the returned `SmpteTimeCode` hours, minutes, seconds & frames will incorrectly be calculated using drop frame arithmetic. These corrupted values will then be passed back into `add_frames()` again as each word is processed and the issue compounds over the line.

For larger timestamps (where the number of "dropped" frames would be larger) this can result in timecodes jumping ahead by many frames each iteration.

I measured this in a REPL environment where `add_frames()` passes the DF flag and `add_frames_buggy()` is a copy with the flag passing removed, then simulated processing a 9 word line (8 add_frames() calls). You can see that the buggy implementation without this fix jumps ahead significantly more than would be expected. 

```python
>>> from fractions import Fraction
>>> from ttconv.time_code import SmpteTimeCode, FPS_29_97
>>> tc = SmpteTimeCode.parse("01:00:00:00", FPS_29_97)
>>> print(float(tc.to_temporal_offset()))
3603.6
>>> tc.add_frames_buggy()
>>> tc.add_frames_buggy()
>>> tc.add_frames_buggy()
>>> tc.add_frames_buggy()
>>> tc.add_frames_buggy()
>>> tc.add_frames_buggy()
>>> tc.add_frames_buggy()
>>> tc.add_frames_buggy()
>>> print(float(tc.to_temporal_offset()))
3632.6957333333335
>>> print(tc, tc.is_drop_frame())
01:00:29:02 False
>>> tc = SmpteTimeCode.parse("01:00:00:00", FPS_29_97)
>>> print(float(tc.to_temporal_offset()))
3603.6
>>> tc.add_frames()
>>> tc.add_frames()
>>> tc.add_frames()
>>> tc.add_frames()
>>> tc.add_frames()
>>> tc.add_frames()
>>> tc.add_frames()
>>> tc.add_frames()
>>> print(float(tc.to_temporal_offset()))
3603.8669333333332
>>> print(tc, tc.is_drop_frame())
01:00:00:08 False
```

Please let me know if you think I've got something wrong here, or if there are adjustments required.

powelli:
Just noticed the mention of unit tests, I think I may have missed updating these, totally valid. I'll update shortly.

palemieux:
@powelli Ok to merge?

powelli:
@palemieux Yes please! 🙂

Issue #509, opened 2026-04-23
SCC Reader configurable framerate


powelli:
## Context

SCC files encode timecodes using a frame-based format, but the current reader assumes a fixed framerate (typically 29.97 fps). SCC (Scenarist Closed Captions) was developed in the early days of digital broadcast, where 29.97 fps NTSC was essentially the only target - so a fixed framerate assumption was baked in from the start.

In practice, SCC files are used with a variety of framerates (23.976, 25, 30, etc.), and there's no way to tell from the file itself which one applies, that context comes from elsewhere such as the video the captions are authored against.

Without configurable framerate support, timecode-to-time conversions can be subtly wrong, causing captions to drift out of sync when the source material isn't 29.97 fps. Support for frame rate configuration was included in the SCC Writer configuration added in #441 

## Proposed changes

- Allow callers to pass either a `frame_rate` (like `SccWriterConfiguration`) or `fps` (like `IMSCWriterConfiguration`) when invoking the SCC reader.
- Retain 29.97 as the default to preserve existing behaviour.
- Drop frame vs non-drop frame will still be inferred from the SCC timecode format, therefore the parameter will not define it (unlike e.g. `29.97NDF` for the SCC writer)
- The provided framerate will be used when converting frame-based timecodes to the internal time-based model

https://github.com/sandflow/ttconv/blob/a426e4e0faef61b07820e4c4098c2129c203491f/src/main/python/ttconv/scc/config.py#L129-L132

https://github.com/sandflow/ttconv/blob/a426e4e0faef61b07820e4c4098c2129c203491f/src/main/python/ttconv/imsc/config.py#L96-L99

powelli:
Branch ready at https://github.com/powelli/ttconv/tree/feat/scc-reader-configurable-framerate - based off changes in review for #506

palemieux:
@powelli I am super not excited for ttconv to support terrible SCC practices. Can you point to widespread use of SCC files that  are not 29.97 DF or NDF?

powelli:
@palemieux - while I agree and am also not super excited by it, I think the fact is that these usages do exist. I don't have first-party information I'm able to share, but do have a direct use-case for this configuration, in which SCC files are being used with matching timecodes for content at 23.976 NDF, 25 & 29.97 DF. Passing these framerates as a configurable parameter on the reader can be demonstrated to solve some specific sync issues encountered with these assets.

[...]

PR #510, opened 2026-04-27
Add SCC Reader support for configurable framerate

powelli:
Addresses https://github.com/sandflow/ttconv/issues/509

It's been noted that using SCC files with non-standard framerates (i.e. those other than 29.97 NDF & 29.97 DF) is questionable practice and easily leads to confusion as there's no direct metadata embedded in the files indicating when this is the case, the original SCC implementation was an encoding of the data from CTA 608 data stream developed for NTSC broadcast at 29.97 fps.

However these files/workflows do occasionally exist in the real world. This change adds support for ingesting these files to enable them to be accurately converted to other formats, but stops short of extending the SCC output with additional non-standard framerates to discourage further use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,15 @@ text alignment.
 
 Default: `"auto"`
 
+#### frame_rate
+
+`"frame_rate" : "29.97" | "30" | "25" | "23.976"`
+
+Specifies the frame rate used to interpret timecodes in the SCC file. Drop-frame vs.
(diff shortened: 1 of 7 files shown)


palemieux:
I have not seen sufficient evidence of interop for SCC files with frame rates other than 29.97 Hz. We can re-open the PR when such evidence becomes available.

PR #518, opened 2026-07-30
 Fix style inheritance in VTT and SRT writers

### LINKED ISSUE #517: SRT/VTT writing: Duplication of inline formatting tags
Converting an SRT or WebVTT cue containing `<i>...</i>` to SRT or WebVTT produces `<i><i>...</i></i>`. 

Repeated conversions keep adding nesting.

(Other styles, such as `<b>`, are affected as well.)

**Minimal repro** 
Input (SRT):
```
1
00:00:01,000 --> 00:00:02,000
<i>Hello</i>
```
Expected: `<i>Hello</i>`
Actual: `<i><i>Hello</i></i>`

**Non-minimal repro**
`input.srt`:
```
1
00:00:01,000 --> 00:00:02,000
<i>Hello, my name is Italics 1</i>

2
00:00:03,000 --> 00:00:04,000
Hello, <i>my</i> name is Italics 2.

3
00:00:05,000 --> 00:00:06,000
<i>Hello, my name is Italics 3.
Second line.</i>
```
Invoked (from fresh cloned directory):
```
PYTHONPATH=src/main/python python3 -m ttconv.tt convert -i input.srt -o output.vtt
```
Resulting `output.vtt`:
```
WEBVTT

1
00:00:01.000 --> 00:00:02.000
<i><i>Hello, my name is Italics 1</i></i>

2
00:00:03.000 --> 00:00:04.000
Hello, <i><i>my</i></i> name is Italics 2.

3
00:00:05.000 --> 00:00:06.000
<i><i>Hello, my name is Italics 3.</i>
<i>Second line.</i></i>
```

**Affected version**
```
❯ git describe --tags
1.2.2-1-g194722a
```
The "problem"(?) can also be seen using the [online-demo](https://ttconv.sandflow.com/) as of 2026-07-30.

palemieux:
* Fix underline style application in VTT and SRT writers

Closes #517

### DIFF
--- a/src/main/python/ttconv/srt/writer.py
+++ b/src/main/python/ttconv/srt/writer.py
@@ -26,7 +26,6 @@
 """SRT writer"""
 
 from __future__ import annotations
-import dataclasses
 import logging
 from fractions import Fraction
 from typing import List, Optional
@@ -81,14 +80,7 @@ class SrtContext:
     self._paragraphs: List[SrtParagraph] = []
     self._text_formatting = config.text_formatting
 
-  @dataclasses.dataclass
-  class _InlineProcessContext:
-    is_bold: bool | None = None
-    is_italic: bool | None = None
-    is_underline: bool | None = None
-    color: str | None = None
-
-  def append_element(self, element: model.ContentElement, begin: Fraction, end: Optional[Fraction], parent_ctx: _InlineProcessContext = None):
+  def append_element(self, element: model.ContentElement, begin: Fraction, end: Optional[Fraction]):
     """Converts model element to SRT content"""
 
     if isinstance(element, model.Div):
@@ -104,7 +96,7 @@ class SrtContext:
       self._paragraphs[-1].set_end(end)
 
       for elem in list(element):
-        self.append_element(elem, begin, end, SrtContext._InlineProcessContext())
+        self.append_element(elem, begin, end)
 
       self._paragraphs[-1].normalize_eol()
 
@@ -112,14 +104,12 @@ class SrtContext:
         LOGGER.debug("Removing empty paragraph.")
         self._paragraphs.pop()
 
-    if isinstance(element, model.Span):
-      is_bold = style.is_element_bold(element)
-      is_italic = style.is_element_italic(element)
-      is_underlined = style.is_element_underlined(element)
-      font_color = style.get_font_color(element)
-
-      # initialize current context to the parent context
-      ctx: SrtContext._InlineProcessContext = dataclasses.replace(parent_ctx) if parent_ctx is not None else SrtContext._InlineProcessContext()
+    if isinstance(element, model.Text):
+      parent_span = element.parent()
+      is_bold = style.is_element_bold(parent_span)
+      is_italic = style.is_element_italic(parent_span)
+      is_underlined = style.is_element_underlined(parent_span)
+      font_color = style.get_font_color(parent_span)
 
       opened_color = False
       opened_bold = False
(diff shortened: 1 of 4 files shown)


hugoheden (approved):
This looks good to me, but some gnarly AI found a very minor issue (see below). 

### (1) What I (as a mere human) would say:

✅ The code now works as expected. (The problematic sample input file now results in nice-looking output SRT and WebVTT.)

✅ The fixed code now also _sanitizes_ input occurrences of `<i><i>`, so that it becomes `<i>`.  Messy example: `<i><b><i>Hello, my name <i>is</i> Italics 1</i></b></i>` becomes: `<i><b>Hello, my name is Italics 1</b></i>` 👍 
(▶️ I might have added a test that shows this behavior)

✅ The tests added look reasonable.
(▶️ I might have added a test that exercises, say, `<b>` as well, to show that the abstraction is genuinely generic)

✅ The added tests would have failed if the fix wasn't introduced.

✅ The fixed code "looks clean to me" (but I am certainly not an expert on that).

✅ (Overall, it's all good for my own purposes, so for egoistical reasons I will mark this as _approved_, but see minor caveat below.)

### (2) Notes from AI

(Used gpt-5.6-terra.)

🤔 Minor: Explicit inner `tts:fontStyle="normal"` style overrides are rendered incorrectly.

(This may be outside the narrow reported regression, but the commit is titled “Fix style inheritance...”, and the new context machinery is intended to model style transitions.)

* Sample TTML input: 
```
<?xml version="1.0"?>
<tt xmlns="http://www.w3.org/ns/ttml"
    xmlns:tts="http://www.w3.org/ns/ttml#styling">
  <body>
    <div>
      <p begin="00:00:01.000" end="00:00:02.000"><span tts:fontStyle="italic">A<span tts:fontStyle="normal">B</span>C</span></p>
    </div>
  </body>
</tt>
```
* **Expected** output:`<i>A</i>B<i>C</i>`
* **Actual** output: `<i>ABC</i>`

palemieux:
@hugoheden See improved processing of nested styles.

hugoheden:
Looks good! At least to me. 

However, "my AI" has some more notes... (Please let me know if this gets frustrating! I am certainly no subtitle expert, so I don't have an opinion of the importance of these findings...)

All of this could be handled as separate issues.

Issue (2) below seems (to me) more important than (1) here. 

---

🤔 (1)  _Minor_: Not fully normalizing/minimizing identical formatting.  

**The output is valid, renders correctly, and importantly - the approach is stable on repeated conversion.** It does not reintroduce the `<i><i>...</i></i>` defect. 

_However_, we don't get fully canonical or minimal output (if that is part of the goal?)

Here is a valid WebVTT snippet that illustrate this - separated into three cases (with inline NOTE-comments):
```
WEBVTT


NOTE 1 - Nested <i> tags.
 Expected: <i>ABC</i>
 Actual:   <i>A</i><i>B</i><i>C</i>

00:00:01.000 --> 00:00:01.500
<i>A<i>B</i>C</i>



NOTE 2 - Nested tags - <i> over <b>
 Expected: <i>A<b>B</b>C</i>
 Actual:   <i>A</i><b><i>B</i></b><i>C</i>

00:00:02.000 --> 00:00:02.500
<i>A<b>B</b>C</i>



NOTE 3 - Adjacent <i> tags
 Expected: <i>ABC</i>
 Actual:   <i>A</i><i>B</i><i>C</i>

00:00:03.000 --> 00:00:03.500
<i>A</i><i>B</i><i>C</i>
```

---

(1b) Addendum: I am not sure how _multiline_ styling should work. Currently, identical formatting is currently split at each cue line break, even though VTT permits one formatting tag to span the newline. Here is some sample input:
```
WEBVTT

00:00:01.000 --> 00:00:01.500
<i>A</i>
<i>B</i>

00:00:02.000 --> 00:00:02.500
<i>A
B</i>
```
**Current behavior:** Both these blocks are converted to separate styling for each cue line (both SRT and WebVTT):
```
<i>A</i>
<i>B</i>
```
**Alternative behavior:** This might also be "correct"?
```
<i>A
B</i>
```
Maybe, however, the _current_ behavior is the _desired_ behavior? Especially for SRT, this behavior feels more conservative (perhaps more safely compatible with various readers...). 

(I mention this, because this behavior could easily accidentally change if the "minimized-output" feature is implemented).

hugoheden:
🤔 (2) Issues with "underline" (`<u>`)

---
(2a): It seems to be that `<u>` is never written (neither for SRT nor WebVTT)? 

```
WEBVTT

NOTE - <u> formatting is lost?
 Expected: <u>ABC</u>
 Actual:   ABC

00:00:01.000 --> 00:00:01.500
<u>ABC</u>
```
I realize that this might be a completely separate issue, but I mention this here so that I can make note (2b) below.

---
(2b): `noUnderline`

Given that (2a) is fixed, the following input - using `noUnderline` - might still be an issue:
```
<tt xmlns="http://www.w3.org/ns/ttml"
    xmlns:tts="http://www.w3.org/ns/ttml#styling">
  <body>
    <div>
      <p begin="00:00:01.000" end="00:00:02.000"><span tts:textDecoration="underline">A<span tts:textDecoration="noUnderline">B</span>C</span></p>
    </div>
  </body>
</tt>
```

Expected output: `<u>A</u>B<u>C</u>`.

Actual output (if 2a is fixed): It is suspected that the `noUnderline` is not properly handled, perhaps resulting in: `<u>ABC</u>`.

palemieux:
Ok. Thanks for finding these long hidden bugs! I think issues (2) is fixed. I do not think issue (1) is worth addressing at this point.

hugoheden:
Nice! And I agree - skip (1).

(Just eyeballing the last commit on my phone here - maybe add the ”underline” tests for SRT as well?)

PR #524, opened 2026-07-31
IMSC reader: accept all(...) syntax in ttp:contentProfiles

### LINKED ISSUE #503: IMSC reader: contentProfiles does not accept the `all(...)` syntax
See https://www.w3.org/TR/ttml2/#profile-attribute-contentProfiles

https://github.com/sandflow/ttconv/blob/a426e4e0faef61b07820e4c4098c2129c203491f/src/main/python/ttconv/imsc/attributes.py#L118

Sanjays2402:
Closes #503

TTML2 allows `ttp:contentProfiles` to carry a combined profile designator list of the form `all(<designator> <designator>)`. `ContentProfilesAttribute.extract()` matched the raw value as a plain whitespace-separated list, so `all(` and `)` were glued onto the first and last designators and `get_content_profiles()` returned mangled URIs. It now recognises the combined form and extracts the designators from inside the parentheses; `any(...)` is logged as unsupported rather than silently mis-parsed, and the plain list form is unchanged.

`test_content_profiles_all_syntax` in `src/test/python/test_imsc_reader.py` fails without the fix and passes with it (the 6 other failures in that file are pre-existing and identical on baseline).

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/src/main/python/ttconv/imsc/attributes.py
+++ b/src/main/python/ttconv/imsc/attributes.py
@@ -117,12 +117,25 @@ class ContentProfilesAttribute:
 
   _CONTENT_PROFILES_RE = re.compile(r"\S+(?:\s+\S+)*")
 
+  # combined profile designator list, e.g. all(<designator> <designator>)
+  _COMBINED_CONTENT_PROFILES_RE = re.compile(r"\s*(all|any)\s*\(\s*(\S+(?:\s+\S+)*?)\s*\)\s*\Z")
+
   @staticmethod
   def extract(ttml_element) -> typing.Optional[typing.Set[str]]:
 
     cr = ttml_element.attrib.get(ContentProfilesAttribute.qn)
 
     if cr is not None:
+
+      m = ContentProfilesAttribute._COMBINED_CONTENT_PROFILES_RE.match(cr)
+
+      if m is not None:
+
+        if m.group(1) == "any":
+          LOGGER.error("ttp:contentProfiles any(...) syntax is not supported")
+          return None
+
+        return set(re.findall(r'\S+', m.group(2)))
 
       m = ContentProfilesAttribute._CONTENT_PROFILES_RE.match(cr)
 
--- a/src/main/python/ttconv/imsc/attributes.py
+++ b/src/main/python/ttconv/imsc/attributes.py
@@ -120,7 +120,7 @@ class ContentProfilesAttribute:
   # combined profile designator list, e.g. all(<designator> <designator>)
   # `all` is the only combinator allowed on ttp:contentProfiles, see
   # https://www.w3.org/TR/2018/REC-ttml2-20181108/#profile-attribute-contentProfiles
-  _COMBINED_CONTENT_PROFILES_RE = re.compile(r"\s*all\s*\(\s*(\S+(?:\s+\S+)*?)\s*\)\s*\Z")
+  _COMBINED_CONTENT_PROFILES_RE = re.compile(r"all\(\s*(\S+(?:\s+\S+)*?)\s*\)")
 
   @staticmethod
   def extract(ttml_element) -> typing.Optional[typing.Set[str]]:
@@ -129,7 +129,7 @@ class ContentProfilesAttribute:
 
     if cr is not None:
 
-      m = ContentProfilesAttribute._COMBINED_CONTENT_PROFILES_RE.match(cr)
+      m = ContentProfilesAttribute._COMBINED_CONTENT_PROFILES_RE.fullmatch(cr)
 
       if m is not None:
 
--- a/src/main/python/ttconv/imsc/attributes.py
+++ b/src/main/python/ttconv/imsc/attributes.py
@@ -118,7 +118,9 @@ class ContentProfilesAttribute:
   _CONTENT_PROFILES_RE = re.compile(r"\S+(?:\s+\S+)*")
 
   # combined profile designator list, e.g. all(<designator> <designator>)
-  _COMBINED_CONTENT_PROFILES_RE = re.compile(r"\s*(all|any)\s*\(\s*(\S+(?:\s+\S+)*?)\s*\)\s*\Z")
+  # `all` is the only combinator allowed on ttp:contentProfiles, see
+  # https://www.w3.org/TR/2018/REC-ttml2-20181108/#profile-attribute-contentProfiles
+  _COMBINED_CONTENT_PROFILES_RE = re.compile(r"\s*all\s*\(\s*(\S+(?:\s+\S+)*?)\s*\)\s*\Z")
 
   @staticmethod
   def extract(ttml_element) -> typing.Optional[typing.Set[str]]:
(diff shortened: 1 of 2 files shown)


palemieux (changes requested):
The model `ContentDocument.set_content_profiles()` supports a set of content profile designator, so the imsc_reader should read the list of designators present in `all(...)` and return a `set`.

Sanjays2402:
That's what it does already, unless I'm misreading you — `extract()` returns a `set` of the designators found inside `all(...)`, which is then handed to `set_content_profiles()` unchanged (`attributes.py` L130-136, called from `elements.py` L154):

```python
m = ContentProfilesAttribute._COMBINED_CONTENT_PROFILES_RE.match(cr)
if m is not None:
    if m.group(1) == "any":
        LOGGER.error("ttp:contentProfiles any(...) syntax is not supported")
        return None
    return set(re.findall(r'\S+', m.group(2)))
```

On the current head (0438acd):

```
'all(http://a http://b)'        -> {'http://a', 'http://b'}
'  all( http://a   http://b )'  -> {'http://a', 'http://b'}
'http://a http://b'             -> {'http://a', 'http://b'}
'any(http://a http://b)'        -> None
```

so the plain list and the `all(...)` list land in the model identically, and `test_content_profiles_all_syntax` covers it. On write, `ContentProfilesAttribute.set()` emits the plain space-separated list, which per the TTML2 definition is the shorthand for the `all(...)` form, so the wrapper isn't needed to round-trip.

The one case that does drop designators is `any(...)`: I return `None` and log, because the model has no way to represent a disjunction and silently turning it into a set would assert conformance to profiles the document only claims one of. If you'd rather keep them and let downstream decide, I'm happy to return the set for `any(...)` too and downgrade the message to a warning — just say which you prefer and I'll push it.

src/main/python/ttconv/imsc/attributes.py
```diff
@@ -117,13 +117,26 @@ class ContentProfilesAttribute:
 
   _CONTENT_PROFILES_RE = re.compile(r"\S+(?:\s+\S+)*")
 
+  # combined profile designator list, e.g. all(<designator> <designator>)
+  _COMBINED_CONTENT_PROFILES_RE = re.compile(r"\s*(all|any)\s*\(\s*(\S+(?:\s+\S+)*?)\s*\)\s*\Z")
```
palemieux: `any()` is not one of the possible keywords for contentProfiles, only for processorProfiles:

https://www.w3.org/TR/2018/REC-ttml2-20181108/#profile-attribute-contentProfiles

src/main/python/ttconv/imsc/attributes.py
```diff
@@ -117,13 +117,26 @@ class ContentProfilesAttribute:
...
 
     cr = ttml_element.attrib.get(ContentProfilesAttribute.qn)
 
     if cr is not None:
 
+      m = ContentProfilesAttribute._COMBINED_CONTENT_PROFILES_RE.match(cr)
+
+      if m is not None:
+
+        if m.group(1) == "any":
```
palemieux: See above. This should never happen for contentProfiles.

palemieux:
>  extract() returns a set of the designators found inside all(...), which is then handed to set_content_profiles() unchanged

+1

> The one case that does drop designators is any(...)

`any()` is not a possible syntax for contentProfiles, only for processorProfiles.

palemieux (changes requested):

src/main/python/ttconv/imsc/attributes.py
Sanjays2402: You are right, `any()` is only defined for ttp:processorProfiles. Dropped it in 64edc64: the regex now matches `all(...)` only, the `any` branch and its log line are gone, and the comment points at the spec section. Anything that is not `all(...)` falls through to the plain-list path as before.

test_imsc_reader.py profile tests pass. The 6 unrelated failures in that file are missing test resource files locally, not the change.

src/main/python/ttconv/imsc/attributes.py
```diff
@@ -117,13 +117,24 @@ class ContentProfilesAttribute:
 
   _CONTENT_PROFILES_RE = re.compile(r"\S+(?:\s+\S+)*")
 
+  # combined profile designator list, e.g. all(<designator> <designator>)
+  # `all` is the only combinator allowed on ttp:contentProfiles, see
+  # https://www.w3.org/TR/2018/REC-ttml2-20181108/#profile-attribute-contentProfiles
+  _COMBINED_CONTENT_PROFILES_RE = re.compile(r"\s*all\s*\(\s*(\S+(?:\s+\S+)*?)\s*\)\s*\Z")
```
palemieux:
```suggestion
  _COMBINED_CONTENT_PROFILES_RE = re.compile(r"all\(\s*(\S+(?:\s+\S+)*?)\s*\)")
```

LWSP are not allowed before and after `all()`.

Prefer to use `fullmatch()` instead of `\Z`.

src/main/python/ttconv/imsc/attributes.py
```diff
@@ -117,13 +117,24 @@ class ContentProfilesAttribute:
...
+
   @staticmethod
   def extract(ttml_element) -> typing.Optional[typing.Set[str]]:
 
     cr = ttml_element.attrib.get(ContentProfilesAttribute.qn)
 
     if cr is not None:
 
+      m = ContentProfilesAttribute._COMBINED_CONTENT_PROFILES_RE.match(cr)
```
palemieux:
```suggestion
      m = ContentProfilesAttribute._COMBINED_CONTENT_PROFILES_RE.fullmatch(cr)
```

palemieux (changes requested):

src/main/python/ttconv/imsc/attributes.py
Sanjays2402: Applied both in d5f4179 - the regex is now `all\(\s*(\S+(?:\s+\S+)*?)\s*\)` with no surrounding LWSP allowed, and the call site uses `fullmatch()` so the `\Z` anchor is gone.

Checked the boundaries: `all(a b)` and `all( a  b )` still match, while ` all(a b)`, `all (a b)` and `all(a b) ` no longer do and fall through to the plain-list path. Both `test_content_profiles` and `test_content_profiles_all_syntax` pass.

palemieux (approved):

palemieux:
@Sanjays2402 Ready to merge?

Sanjays2402:
Yes, ready from my side.

Head is d5f4179, with both of your last suggestions applied: the regex is `all\(\s*(\S+(?:\s+\S+)*?)\s*\)` so no LWSP is allowed around `all()`, and the call site uses `fullmatch()`. Anything that is not exactly `all(...)` falls through to the plain-list path unchanged, and the `any(...)` branch is gone since it is only valid for `ttp:processorProfiles`.

I pulled the `imsc-tests` submodule and re-ran the suite on this head: `test_imsc_reader.py` is 23 passed (the failures I mentioned earlier were just the missing submodule on my side), `test_imsc_writer.py` 43 passed, and the rest of `src/test/python` is green apart from `test_scc_dump.py::test_convert_input_file_type_scc`, which fails on a missing local `.scc` resource and is unrelated to this change.

PR #534, opened 2026-08-16
Do not crash on empty WebVTT input or empty SRT cue text

eeshsaxena:
A couple of the subtitle readers raise on trivial input they otherwise tolerate.

`vtt.to_model` on an empty document crashes:

```python
import io, ttconv.vtt.reader as vtt
vtt.to_model(io.StringIO(""))   # AttributeError: 'NoneType' object has no attribute 'startswith'
```

An empty file leaves only the `None` terminator, and the `START` state calls `line.startswith("WEBVTT")` without the `None` check the other states have. Separately, `srt.to_model` hits `UnboundLocalError` on a cue whose text body is empty (a time line followed immediately by a blank line), since `subtitle_text` is only initialized once a non-empty text line is seen.

I guarded the `None` line in the WebVTT `START` state and initialized `subtitle_text` when the paragraph is created. Added a test for each; the srt and vtt reader suites pass.

### DIFF
--- a/src/main/python/ttconv/srt/reader.py
+++ b/src/main/python/ttconv/srt/reader.py
@@ -270,6 +270,7 @@ def to_model(data_file: typing.IO, _config: SRTReaderConfiguration = None, progr
         return None
 
       current_p = model.P(doc)
+      subtitle_text = ""
 
       current_p.set_begin(
         int(m.group('begin_h')) * 3600 + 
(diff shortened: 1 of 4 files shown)


palemieux:
@eeshsaxena Thanks for raising these parsing issues. It looks like the VTT reader never handled VTT files with incorrect signatures according to the [specification](https://www.w3.org/TR/webvtt1/#file-parsing). I have proposed a fix at:

https://github.com/sandflow/ttconv/pull/536

SRT files will be tackled in a separate PR.

I would very much appreciate your review.

eeshsaxena:
Thanks for taking this on. Your signature-check approach in #536 is better than my guard, happy to see it handled per the spec. Left one note there about the `None` return needing handling at the `tt.py` call site. Fine to close this in favour of #536.

eeshsaxena:
Closing in favor of #536, which takes a more complete approach to the WebVTT signature handling (proper spec-based first-line check, returning `None` on a bad signature, and broader test coverage). I left a review there with two small notes. Thanks @palemieux for picking this up.

PR #535, opened 2026-08-16
Ignore unmatched end tags in SRT text

eeshsaxena:
An SRT text line that contains an end tag with no matching start tag crashes:

```python
import io, ttconv.srt.reader as srt
srt.to_model(io.StringIO("1\n00:00:01,000 --> 00:00:04,000\n</b>text\n"))
# TypeError: Children of body must be P instances
```

`_TextParser.handle_endtag` walks `self.parent` up a level unconditionally, so a stray `</b>` moves the insertion point above the paragraph and the text that follows gets pushed onto the div or body, which only accept `P` children. I made it only move up when the current parent is actually a span it opened, which just ignores the unmatched tag. Added a test; the srt reader suite passes.

### DIFF
--- a/src/main/python/ttconv/srt/reader.py
+++ b/src/main/python/ttconv/srt/reader.py
@@ -179,7 +179,11 @@ class _TextParser(HTMLParser):
       return
 
   def handle_endtag(self, tag):
-    self.parent = self.parent.parent()
+    # Ignore an unmatched end tag. Without this guard a stray "</b>" would
+    # walk the parent above the paragraph, and following text would be pushed
+    # onto the div or body, which only accept P children.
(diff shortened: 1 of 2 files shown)


palemieux:
@eeshsaxena Thanks for the bug report. I propose the following, which adds error messages:

https://github.com/sandflow/ttconv/pull/539/changes

eeshsaxena:
Thanks, adding proper error messages in #539 works for me. My change just stopped the crash, yours is the cleaner treatment. Happy to close this in favour of it.

palemieux:
Replaced with https://github.com/sandflow/ttconv/pull/539

PR #541, opened 2026-08-17
srt_reader: handle empty cues

### LINKED ISSUE #540: SRT Reader dies on empty cues
SRT Reader dies on empty cues:

```
1
101:00:00,000 --> 101:00:01,000

2
101:00:00,000 --> 101:00:01,000
Bonjour
```

From https://github.com/sandflow/ttconv/pull/534/changes

palemieux:
Closes #540


palemieux:
@eeshsaxena I would appreciate your review.

PR #548, opened 2026-09-01
scc/writer: delay colliding pop-on captions instead of dropping them

jason-valenzuela:
## Problem

`scc_writer.from_model` silently drops whole pop-on captions when a caption's line-21 packets cannot be transmitted back-to-back with the previous caption's (the `Skipping ISD … due to overlap in line 21 packets` path). On dense content this loses caption text with only a log warning.

## Fix

When a pop-on caption collides with the previous one, delay its onset to sit right after the previous caption's packets instead of dropping it, carrying the preceding erase (EDM) into the delayed chunk so no stale erase wipes it and the emitted timecodes stay monotonic and non-overlapping. The pull-back of the preceding erase is bounded to the minimum needed, so the previous caption keeps as much of its authored on-screen time as possible (and an author-written short cue is never itself penalised). A caption that still cannot achieve a minimum readable window even after delaying is dropped honestly (the existing warning) rather than emitted as a two-frame flash.

The minimum readable window is exposed as a new `SccWriterConfiguration` field `min_popon_display_frames` (default `15` ≈ 0.5 s at 29.97 fps; `0` disables dropping entirely, preserving maximal text at the cost of possible flashes). The common, non-colliding path is unchanged — existing golden-SCC tests are untouched.

## Tests

- `test_dense_popon_no_line_drop` — dense captions previously dropped now survive a write→read round-trip (exact SCC asserted)
- `test_dense_popon_emitted_scc_is_ordered` — emitted timecodes monotonic and non-overlapping
- `test_dense_popon_cascade_all_emitted_captions_displayable` — every emitted caption meets the display floor
- `test_dense_popon_authored_short_prev_not_dropped` — a short authored cue no longer forces an unnecessary drop
- `test_dense_popon_drop_only_when_infeasible` — a caption is dropped only when no readable window exists
- `test_dense_popon_erase_pullback` — the bounded erase pull-back path (erase moved minimally, both captions keep a window above the floor)
- `test_min_popon_display_frames` — config parse/default/validation

Full suite: 617 passed, 0 failures, 0 errors, 6 pre-existing skips.

## Notes

- Unrelated pre-existing bug noticed nearby (not addressed here): `from_model` indexes `chunks[-2]` under a `len(chunks) > 0` guard, which raises `IndexError` when exactly one chunk exists — happy to file a separate issue.


palemieux:
@jason-valenzuela I would be tempted to wait until all chunks are collected before moving them.

jason-valenzuela:
> @jason-valenzuela I would be tempted to wait until all chunks are collected before moving them.

Got it! Will adjust my approach here.

jason-valenzuela:
@palemieux Updated the approach here, let me know what you think! The chunk-moving logic is now run after the all the pop-on chunks are collected.

src/main/python/ttconv/scc/writer.py
```diff
@@ -321,6 +321,10 @@ def _isd_progress(progress: float):
     captions.append(caption)
 
   chunks : List[_Chunk] = []
+  # pop-on captions are collected here at their natural (undelayed) positions and
+  # line 21 collisions are resolved in a post-pass below, once the whole list is
+  # known: (enm chunk, its erase chunk or None, caption begin time)
+  popon_captions: List = []
```
palemieux: Why create a popon_captions list when chunks could be sufficient?

src/main/python/ttconv/scc/config.py
```diff
@@ -107,6 +107,14 @@ def _decode_rollup_lines(value: str) -> int:
 
   return decoded_value
 
+def _decode_min_popon_display_frames(value: str) -> int:
```
palemieux: I am not sure there is every a correct value here since the duration of a caption is related to the amount of text present. I would define a simply boolean (by default true) that indicates whether overlapping captions should be dropped.

palemieux (commented):
Is the issue happening with content that was originally authored for 608, or content that was authored in other formats?

jason-valenzuela:
> Is the issue happening with content that was originally authored for 608, or content that was authored in other formats?

This issue is happening for content that was authored in other formats! I've been seeing this issue when we're doing VTT -> SCC conversions.

src/main/python/ttconv/scc/config.py
jason-valenzuela: Got it, will switch this over to a boolean config!

src/main/python/ttconv/scc/writer.py
jason-valenzuela: I added the intermediate list to keep each caption's (paint, erase) pair and its begin explicit for the delay pass and the warning. But I can build the paint/erase chunks straight into chunks in the loop and run the collision pass over chunks directly, pairing them there, and remove the extra list.

palemieux:
> This issue is happening for content that was authored in other formats! I've been seeing this issue when we're doing VTT -> SCC conversions.

Are the unit test files, e.g., `src/test/resources/scc/vtt/dense_popon_cascade.vtt`, representative of actual VTT content you have seen, or just created to ensure overlap?

jason-valenzuela:
> > This issue is happening for content that was authored in other formats! I've been seeing this issue when we're doing VTT -> SCC conversions.
> 
> Are the unit test files, e.g., `src/test/resources/scc/vtt/dense_popon_cascade.vtt`, representative of actual VTT content you have seen, or just created to ensure overlap?

These are excerpts from real captions we've processed in production.

palemieux:
This is super useful information.

```
1
00:00:30.000 --> 00:00:33.561
Alvarez's perspective, I am done with this project. I'm

2
00:00:33.561 --> 00:00:37.120
ready for something else. Do you think London would be enticing enough

3
00:00:37.120 --> 00:00:37.481
for him?

4
00:00:38.640 --> 00:00:42.091
Let me just pick up on something that you mentioned there, why Atletico Madrid

```

Looking at the timing (no gaps between cues) and the text itself (e.g., breaking after "I'm"), I would be surprised if these were authored in VTT. Perhaps they were (badly) converted from 608 roll-up captions?

The basic issues is that 608 pop-on captions are double-buffered: the time that you have to write caption N to the back-buffer is the on-screen duration of caption N-1. In the case above, the duration of 3 is not sufficient to buffer 4.

Could we consider fixing the content at the model, before conversion to SCC. Could we, for instance, look into merging temporally adjacent cues and reflowing the text and using roll-up instead of pop-on?

palemieux:
@jason-valenzuela P.S.: You can find me at https://discord.gg/mYpGjdwMH

jason-valenzuela:
Thanks for that invite, just joined!
Honestly I'm a bit stumped on the best approach here; the merge+reflow approach could definitely work, but I'm hesitant to go that route since it changes the segmentation and wrapping that the source chose. After going through more of our prod traffic, I'm seeing our issues fall into two buckets:

1. Pop-on captions that are just densely timed. Distinct two-line captions, no scroll carry-over:

```
00:00:01.600 --> 00:00:04.080
This is a joint statement
by the United Women's News

00:00:04.080 --> 00:00:06.320
Conference to announce
action in response

00:00:06.320 --> 00:00:09.720
to the abortion situation
in New York state.
```
That's what this PR handles, by delaying instead of dropping.

2. Roll-up content where the roll-up detector returns false partway through. `is_rollup` (in `src/main/python/ttconv/scc/writer.py`) is global and sticky to `False`, so one caption that doesn't continue the scroll flips the whole rest of the document to pop-on:

```
00:00:50.751 --> 00:00:53.952
ENTERTAINMENT AND POLITICS TO
TALK ABOUT THE FAILURES,

00:00:54.754 --> 00:00:58.991
SETBACKS AND DEFINING MOMENTS.
HERE'S KENDRA SCOTT HOW SHE

00:00:58.993 --> 00:01:00.259
HERE'S KENDRA SCOTT HOW SHE
LEARNED SHOWING KINDNESS IS NOT
```

`SETBACKS AND DEFINING MOMENTS.` doesn't scroll from `TALK ABOUT THE FAILURES,`, so the detector returns the whole document as pop-on.

I'm thinking that the second issue could be handled separately, and we can allow the detector to read the whole document before determining if it's roll-up or pop-on. I can handle that part if that sounds good to you.

Also, I can send the full VTT sources from these examples over Discord if that would help!

palemieux:
Re (1), I am thinking of the following re-flow to roll-up, which preserves line breaks:

```
00:00:01.600 --> 00:00:02.600
This is a joint statement

00:00:02.600 --> 00:00:03.600
This is a joint statement
by the United Women's News

00:00:03.600 --> 00:00:04.600
This is a joint statement
by the United Women's News
Conference to announce

00:00:04.600 --> 00:00:05.600
by the United Women's News
Conference to announce
action in response

00:00:05.600 --> 00:00:06.600
Conference to announce
action in response
to the abortion situation

00:00:06.600 --> 00:00:07.600
action in response
to the abortion situation
in New York state.

00:00:07.600 --> 00:00:08.600
to the abortion situation
in New York state.

00:00:08.600 --> 00:00:09.720
in New York state.
```

Zero temporal gap between successive pop-on captions suggest that the tool that created them has a bug, i.e., it incorrectly handled roll-up captions.

palemieux:
> I'm thinking that the second issue could be handled separately, and we can allow the detector to read the whole document before determining if it's roll-up or pop-on. I can handle that part if that sounds good to you.

Ok. Yes, this a separate issue, which I filed under https://github.com/sandflow/ttconv/issues/549

PR #550, opened 2026-09-04
SCC writer: improve roll-up detection

### LINKED ISSUE #549: Roll-up detection fails
From @jason-valenzuela at https://github.com/sandflow/ttconv/pull/548#issuecomment-5541092650

Roll-up content where the roll-up detector returns false partway through. is_rollup (in src/main/python/ttconv/scc/writer.py) is global and sticky to False, so one caption that doesn't continue the scroll flips the whole rest of the document to pop-on:

```
00:00:50.751 --> 00:00:53.952
ENTERTAINMENT AND POLITICS TO
TALK ABOUT THE FAILURES,

00:00:54.754 --> 00:00:58.991
SETBACKS AND DEFINING MOMENTS.
HERE'S KENDRA SCOTT HOW SHE

00:00:58.993 --> 00:01:00.259
HERE'S KENDRA SCOTT HOW SHE
LEARNED SHOWING KINDNESS IS NOT
SETBACKS AND DEFINING MOMENTS. doesn't scroll from TALK ABOUT THE FAILURES,, so the detector returns the whole document as pop-on.
```

I'm thinking that the second issue could be handled separately, and we can allow the detector to read the whole document before determining if it's roll-up or pop-on. I can handle that part if that sounds good to you.

palemieux:
Closes #549


palemieux:
@jason-valenzuela Let me know what you think!

src/main/python/ttconv/scc/writer.py
```diff
@@ -308,18 +303,22 @@ def _isd_progress(progress: float):
...
+  for i in range(1, len(captions)):
+    # do not detect roll-up if successive captions are more than 1 frame apart
+    # ideally we would ignore successive captions unless they are contiguous,
+    # but many tools incorrectly set end times to be inclusive instead of exclusive
+    if abs(captions[i - 1].get_end() - captions[i].get_begin()) > Fraction(1, 30):
+      continue
+
+    if captions[i][-1].startswith(captions[i - 1][-1]) or \
+      len(captions[i]) > 1 and captions[i][-2] == captions[i - 1][-1]:
+      is_rollup = True
```
jason-valenzuela: Been running this branch against more of our prod traffic and I think the detection is a bit too eager here. The first pair that continues the scroll sets `is_rollup` for the whole document and breaks, so a genuine pop-on file with just a couple of incidental repeats flips to roll-up and starts losing text.

Here's one I'm seeing: genuine pop-on, 131 cue pairs, and only 2 of them "match," each just a one-word line that happens to recur in the next cue:

```
00:02:02.540 --> 00:02:03.420
Lisbon.

00:02:03.420 --> 00:02:04.360
Lisbon.

00:04:02.000 --> 00:04:02.500
Huh?

00:04:02.500 --> 00:04:03.160
Huh?
```

Both of those are 0 ms apart so the new 1-frame gap check doesn't filter them, and either one on its own is enough to flip the whole file. On master this stays pop-on; on this branch it comes out roll-up and drops 55 lines.

I'm wondering if we could let the detector look at the whole document and go by a fraction of continuations rather than any single match, something like `continuations / (len(captions) - 1) >= 0.5`.

src/main/python/ttconv/scc/writer.py
```diff
@@ -308,18 +303,22 @@ def _isd_progress(progress: float):
...
+      break
+
   chunks : List[_Chunk] = []
   for i, caption in enumerate(captions):
```
jason-valenzuela: One more thing I ran into on the emission side: in the roll-up branch of this loop we only ever write `caption[-1]` (the emission a few lines down, ~line 353), so any time the top line is actually new (the first caption, or a hard reset where nothing carries forward from the previous one) that line just never gets emitted.

Here's a real one from our prod traffic. This file goes pop-on on master and roll-up on this branch, and the opening caption's top line disappears:

```
00:00:08.908 --> 00:00:11.176
>>> KIND, WHEN HE WAS BORN, WE
WERE IN A SITUATION NOT SIMILAR
```

On master (pop-on) both lines survive; on this branch (roll-up) `>>> KIND, WHEN HE WAS BORN, WE` is dropped, since the first caption isn't a paint-on so only its last line is written.

I think we'd want to treat a caption that neither paints-on nor scrolls as a reset: EDM to clear, then emit every line (a `CR` + PAC per line) instead of just the last one.

jason-valenzuela (commented):
Took this branch for a spin against our prod VTT->SCC traffic and ran into two spots where caption text gets dropped. Details inline. I'll send the caption files I'm referencing over separately so you can repro on your end.

src/main/python/ttconv/scc/writer.py
palemieux: Should be fixed by https://github.com/sandflow/ttconv/pull/550/commits/5891e9406b262b42394c4c0ed4073fb0ba9a1aac

src/main/python/ttconv/scc/writer.py
palemieux: Should be fixed at https://github.com/sandflow/ttconv/pull/550/commits/f5d26d79dc588280d8cf1dfcc72bf0fd1ba33f91

jason-valenzuela (approved):
Pulled the latest and re-ran it against our most recent prod VTT→SCC traffic and all issues have been resolved!

